PR #7, opened 2018-07-04
Modify build script to zip with version number

herman-rogers:


lanwen (approved):
next step to use npm wrappers for tar, cp and npm itself? :)
 Just kidding :) IMO now it looks more accurate 👍 

But what the actual reason?

PR #10, opened 2018-08-14
Update.build.process

herman-rogers:
- Move most heavy lifting to circle ci config
- Add git tagging to build.js
- remove old build process
- clean up unnecessary files


herman-rogers:
Don't be scared by the size of the diff. Most of it is static website cruft that has been remove (and it is not expanded in the diff view).

.circleci/config.yml
```diff
@@ -40,7 +45,36 @@ jobs:
 
       - run: npm install
       - run: git config --global push.default simple
-      - run: npm run release
+      - run: npm run build:package
+      - run: cd ./build
+      - run: echo "//registry.npmjs.org/:_authToken=$npm_TOKEN" > ~/repo/.npmrc
+      - run: npm publish --access public
```
lanwen: can we just grab it directly from github? what for do we need npm?

package.json
```diff
@@ -1,16 +1,22 @@
 {
-  "name": "vivy-components",
-  "version": "0.0.5",
-  "private": true,
+  "name": "@vivy/vivy-components",
+  "version": "0.0.6",
```
lanwen: should we update it automatically?

.circleci/config.yml
lanwen: with npm we can't unpublish it easily

package.json
herman-rogers: We'll have to update it manually anyways because even if we did increment it we'd need to commit the changes to git which will then retrigger master and cause an infinite loop.

.circleci/config.yml
```diff
@@ -40,7 +45,36 @@ jobs:
...
+          - v1-dependencies-{{ checksum "package.json" }}
+          # fallback to using the latest cache if no exact match is found
+          - v1-dependencies-
+
+      - run: npm install
+      - run: git config --global push.default simple
+      - run: npm run build:storybook
+      - run: mkdir .static-website
+      - run: cd .static-website
+      - run: git clone -b gh-pages git@github.com:UvitaTeam/vivy-components .
```
lanwen: is it better to just force push to gh-pages?

.circleci/config.yml
herman-rogers: Sure but we can contact npm to take it down if we ever cross that bridge. Worse case we just deprecate it.

package.json
lanwen: mb filter it with commit message or trigger explicitly with tag?

.circleci/config.yml
```diff
@@ -16,15 +16,20 @@ jobs:
 
       - run: npm install
       - run: npm run linter:check
-      - run: npm run release:test
-      - run: node build.js '0.0.1'
+      - run: npm run build:package
+      - run: npm run build:storybook
+      # check npm token exists
+      - run: echo "//registry.npmjs.org/:_authToken=$npm_TOKEN" > ~/repo/.npmrc
+      - run: cat ~/repo/.npmrc
```
lanwen: better to not leak token in this way :) can it be even not saved at all?

.circleci/config.yml
herman-rogers: This way is a bit more work but less destructive.

.circleci/config.yml
herman-rogers: Hmm I can remove it no problem but is it really a leak? The output afaik only get's outputted into circle ci which requires auth to view. I mean really someone could ssh into the container and just cat out the credentials. However, we do need to save it so we can publish.

package.json
herman-rogers: Sorry, I don't follow

.circleci/config.yml
lanwen: logs usually available and viewable for oss projects. But even with auth - we should not show token to everyone with gh access from the early days. Otherwise, we should reset it every time someone got fired :)

.circleci/config.yml
```diff
@@ -16,15 +16,19 @@ jobs:
 
       - run: npm install
       - run: npm run linter:check
-      - run: npm run release:test
-      - run: node build.js '0.0.1'
+      - run: npm run build:package
+      - run: npm run build:storybook
+      # check npm token exists
+      - run: echo "//registry.npmjs.org/:_authToken=$npm_TOKEN" > ~/repo/.npmrc
```
lanwen: found that npm actually can read from the environment variables to interpolate this string by himself.
so better to do it as `echo '//registry.npmjs.org/:_authToken=${npm_TOKEN}' > ~/repo/.npmrc` (with single quotes)

package.json
herman-rogers: We might be able to skip builds based on commit message - but if we don’t tag the build (or increment the package) the build itself will fail and not be published. We can check if it’s been changed but we’ll have to do it in code.

.circleci/config.yml
herman-rogers: Except that circle will inject the token and it does in fact insert the token and not a variable. I was double checking this was true with the cat command (because I also was skeptical). If we expose it in brackets will that cause any sort of leak?

package.json
lanwen: what I'm worrying about is that we can forget to change the version here, or it can be inconvenient to do it manually as requires commit, where tag can be created using github release interface

.circleci/config.yml
lanwen: wut? I can't understand you.

- in build you have environment variables
- they can be expanded by shell or by npm
- with shell you will persist the token - and this is bad as it can be cached, or someone could have access to the fs
- so if npm can expand by himself this variable, why ever save it expanded?
- the curly brackets is how actually env variables should be written if they are part of the sting (code-style)

herman-rogers: I’ll come to you to discuss but I think I get what you’re saying. I got lost in the last bit my bad.

lanwen (approved):

PR #11, opened 2018-08-17
Add.icons.and.buttons

herman-rogers:


lanwen:
are we sure we want to put our icons under MIT license?

herman-rogers:
I don't think it's a big deal. It's not like we can't reverse it if necessary but there probably isn't a lot of IP tied up into the icons.

lanwen:
I think we should add a note that these icons should have at least credentials

herman-rogers:
These icons should have credentials? Like, add a copyright to them or something?

lanwen:
Exactly

herman-rogers:
Added

lanwen (approved):

PR #13, opened 2018-09-07
adds.semantic.ui

paschalidi:
a PR for adding react.semantic-ui to vivy-components lib


lanwen:
why do we ever need this? semantic ui is a completely independent thing and if you need something - you should adopt it

lanwen:
thats completely wrong way to do we have own components library to avoid dependency on any other frameworks and use only self-described components please discuss such changes on web dev sync or at least at web-dev channel before even start

PR #15, opened 2018-09-10
imports different reusable components.

paschalidi:


stories/footer.story.jsx
```diff
@@ -0,0 +1,35 @@
...
+  width: "80%",
+  margin: "0 auto",
+  paddingTop: "100px"
+};
+
+storiesOf("Footer", module).addWithJSX("Simple Vivy Footer", () => (
+  <div style={container}>
+    <Index>
+      <Fragment>
+        <a className="link" href="https://www.vivy.com/impressum" target="_">
```
herman-rogers: this is my bad, but when we use target="_" i.e. blank we should also add noopener https://github.com/asciidoctor/asciidoctor/issues/2071

paschalidi: yes agree. I just didn't mean to change anything you had there. I am doing the changes as we speak

PR #16, opened 2018-09-10
Syncs rules on eslint-prettier also styles the components accordingly

paschalidi:
after this is approved both projects emergency-data and vivy-components will listen to the same rules


lanwen (approved):

.eslintrc.json
```diff
@@ -1,21 +1,29 @@
...
     "document": false,
     "URLSearchParams": false,
     "fetch": false,
-    "it": false
-  },
-  "rules": {
-    "prettier/prettier": [
-      "error",
-      {
-        "singleQuote": true,
```
lanwen: btw why from single quotes to double ones?

paschalidi: So that the rules between eslint and prettier are the same. 
Webstorm is complaining a lot and since single-quotes double-quotes are not really important I thought of changing that. Plus there is no real reason adding this rule there in the first place. 
If you find that this is not nice then we can chat about it.

PR #21, opened 2018-09-12
MAIN Add layout, collections, and new icons

herman-rogers:


src/MedicalCard/dateFormat.js
```diff
@@ -0,0 +1,21 @@
+const monthNames = [
```
lanwen: should it be prepared for localization?

src/MedicalCard/detailedIcon.style.js
```diff
@@ -0,0 +1,31 @@
...
+    font-size: 26px;
+    color: #575756;
+  }
+
+  .icon-detail {
+    text-align: center;
+    font-size: 20px;
+    font-weight: 400;
+    line-height: 1;
+    color: #575756;
```
lanwen: i see a lot of colors inlined - should they be organized like a pallet in one place?

stories/simpleLoader.story.jsx
```diff
@@ -1,23 +1,13 @@
...
-storiesOf("Animations", module).addWithJSX(
+storiesOf("Animations", module).add(
   "Spinner",
-  withInfo("A Spinner")(() => (
-    <div style={container}>
-      <SimpleLoader />
-      <SimpleLoader isCentered />
+  withInfo("A simple spinner for showing loading progress")(() => (
+    <div>
+      <Spinner isCentered />
```
lanwen: why not just `centered`?

src/MedicalCard/detailedIcon.style.js
herman-rogers: So an inlined color would be, like 
```html
<div style={color: #23432} </div>
```

So I'm slightly confused but I'm going to interpret this as, why don't we have one location where we import the colors (like a palette, or a custom theming thing). And if that's the case then its a good point. Probably should be its own thing but Christos has also mentioned that before so might be an idea.

src/MedicalCard/dateFormat.js
herman-rogers: Cool so translations are next on my todo list after some minor style changes in the emergency service. I’ll start here for translations with another PR to make these general.

lanwen (approved):

PR #23, opened 2018-09-18
component-BirthDate: imports BirthDate component

paschalidi:


src/BirthDate/BirthDate.jsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+export default function BirthDate({ birthDate, style }) {
```
lanwen: just date?

stories/index.stories.js
```diff
@@ -7,5 +7,6 @@ import "./grid.story";
 import "./button.story";
 import "./forms.story";
 import "./layout.story";
+import "./patientBirthDate.story";
```
herman-rogers: Does this need its own story or can we add it to a subsection? For example, PatientData (or Display or whatever) and under that we have birthDate - we could also add MedicalID under the Patient story.

paschalidi: Tbh, for me every Component must have its own stories. 
After looking into different examples and trying to figure out the best way to structure a `storybook` this is what I got out of it: 

Every `component` has his own `story` and the `story` is marked with the component name so that we can easily find the component when we visit the storybook website that we generate (on the left side menu it must be the names of the components so we can easily go and find em). 

Every component `story` has many different subsections regarding the component `props` and definitions `<Button primary>Press</Button>` would be one substory of `Button` and `<Button secondary>Press Sec</Button>`   would be another sunstory. 

https://github.com/bufferapp/buffer-components
https://github.com/quran/common-components/blob/master/src/Button/button.stories.js
https://github.com/IBM/carbon-components-react/blob/master/src/components/Accordion/Accordion-story.js

Here you can find some examples of what I mean. 
I know we are trying to structure them regarding the business model but in a storybook I dont think this is makes much of a sense, (there would be many times that a component will be shared between different parts of the app)

We can talk about it on the web-sync or just something to think about.

I wouldnt mind moving it to a subcategory if you point one but I believe building a component library is something that has no relation with the business logic as we try to separate the concept of components and business logic and not to cary the complexity of the business to this package as well.

herman-rogers: Fair enough, so should there be a story for Dates and we put BirthDate underneath it? Also, should we not call this patientBirthDate but rather just birthDate. I'm all for removing the business logic out of the components.

herman-rogers: Or..are you saying we should just have BirthDate / SimpleDate / Button

paschalidi: > Also, should we not call this patientBirthDate but rather just birthDate
this is my bad

herman-rogers (approved):

stories/birthDate.story.jsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import { withInfo } from "@storybook/addon-info";
+import BirthDate from "../src/BirthDate";
+
+storiesOf("BirthDate", module).add(
+  "BirthDate",
+  withInfo("Simply shows the birthdate in form DD.MMM.YYYY")(() => (
```
herman-rogers: Oh wait small one, should this be DD.MM.YYYY? Or is it in the format Aug, Dec ?

paschalidi: We use this format in 3 places as far as I know. I dont know what is nicer having two separate components with different format or one and we parse a prop on it and regarding this prop we show different format? 
Dont have a strong opinion here

paschalidi: if you see the comments Madison asked for it

stories/index.stories.js
paschalidi: I would just say BirthDate / SimpleDate / Button
This was we have 2 pros.
1. we never argue again on which category a new component should be put uner
2. uniformity on the local storybook website. With one look you could see all the components we have and opening a directory would show you how many different options you have on how to use the specific component.

stories/birthDate.story.jsx
herman-rogers: Cool cool just curious

PR #26, opened 2018-09-21
fixes-birthDate-component: fixes bug with component not return something

paschalidi:


herman-rogers (approved):

src/BirthDate/BirthDate.jsx
```diff
@@ -2,12 +2,11 @@ import React from "react";
...
 export default function BirthDate({ date, style }) {
-  if (date)
-    return (
-      <div style={style}>
-        {date.split("-")[2]}.{date.split("-")[1]}.{date.split("-")[0]}
-      </div>
-    );
+  return (
+    <div style={style}>
+      {date.split("-")[2]}.{date.split("-")[1]}.{date.split("-")[0]}
```
lanwen: will it panic in case of date is undefined?

herman-rogers: It'll complain in the proptypes since it is required.

lanwen: but shouldn't we handle it gracefully?

herman-rogers: Well we could do it, but I don't think its up to the component creator but rather the people using the component. Other wise all of our components will have {exist ? prop : handle gracefully} which doesn't really scale or work. I don't think its unreasonable to require a user to use a component properly.

lanwen (approved):

PR #28, opened 2018-09-25
Exports SimpleLoader

paschalidi:
* exports SimpleLoader
* deletes duplicated component


paschalidi:
Its confusing but this component was existent 2 times. 
Was never been exported tho.
So it is not used in any of our apps, now it will be exported so that can be used instead of the `semantic-ui` `Loader` on the `uvita-webapp`

stories/simpleLoader.story.jsx
```diff
@@ -1,13 +1,13 @@
...
 
 storiesOf("Animations", module).add(
-  "Spinner",
-  withInfo("A simple spinner for showing loading progress")(() => (
+  "SimpleLoader",
+  withInfo("A simple loader")(() => (
     <div>
-      <Spinner centered />
+      <SimpleLoader centered />
     </div>
```
lanwen: Can you add different sizes to story?

lanwen (approved):

stories/simpleLoader.story.jsx
paschalidi: yes I will.

herman-rogers (approved):

PR #29, opened 2018-09-27
MAIN - updates-icons: updates icons with chevrons

paschalidi:
* updates icons with 
    * chevron-up
    * chevron-down
    * chevron-left
    * chevron-right


public/fonts/vivy-icons/iconMapping.json
```diff
@@ -17,5 +17,9 @@
   "icon-will-information": "e80e",
   "icon-height-icon": "e801",
   "icon-insurance-shield": "e810",
-  "icon-phone-number": "e811"
+  "icon-phone-number": "e811",
```
herman-rogers: looks like e811 conflicts also with "phone-number" above, should phone-number just be removed and this used instead?

paschalidi: like this?

herman-rogers (approved):

PR #30, opened 2018-09-27
Release.0.4.0

paschalidi:
image
* updates button so that it could be either anchor, so that it can open links, download documents etc
* exports component footer
* adds style to Icon component so that when we want to change color and weight or font-size we parse prop `style` to the component itself

### DIFF
--- a/src/button/AnchorButton.jsx
+++ b/src/button/AnchorButton.jsx
@@ -13,13 +13,13 @@ export default function AnchorButton({
   return (
     <ButtonStyles>
       <a
-        onClick={onClick}
         className={type}
+        onClick={onClick}
         href={href}
         target={target}
         rel={rel}
       >
-        {children}
+        <div className="anchor-child">{children}</div>
       </a>
     </ButtonStyles>
   );
--- a/src/button/AnchorButton.jsx
+++ b/src/button/AnchorButton.jsx
@@ -1,0 +1,42 @@
+import React from "react";
+import PropTypes from "prop-types";
+import ButtonStyles from "./button.style";
+
+export default function AnchorButton({
+  onClick,
+  type,
+  href,
+  target,
+  rel,
+  children
+}) {
+  return (
+    <ButtonStyles>
+      <a
+        onClick={onClick}
+        className={type}
+        href={href}
+        target={target}
+        rel={rel}
+      >
+        {children}
+      </a>
+    </ButtonStyles>
+  );
+}
+
+AnchorButton.propTypes = {
+  children: PropTypes.node.isRequired,
+  href: PropTypes.string.isRequired,
+  type: PropTypes.string,
+  onClick: PropTypes.func,
+  target: PropTypes.string,
+  rel: PropTypes.string
+};
+
+AnchorButton.defaultProps = {
+  type: "primary",
+  onClick: () => {},
+  target: "_self",
+  rel: ""
+};
--- a/src/button/AnchorButton.jsx
+++ b/src/button/AnchorButton.jsx
@@ -8,7 +8,8 @@ export default function AnchorButton({
   href,
   target,
   rel,
-  children
+  children,
+  customStyle
 }) {
   return (
     <ButtonStyles>
@@ -18,6 +19,7 @@ export default function AnchorButton({
         href={href}
         target={target}
         rel={rel}
+        style={customStyle}
       >
         <div className="anchor-child">{children}</div>
       </a>
@@ -31,12 +33,14 @@ AnchorButton.propTypes = {
   type: PropTypes.string,
   onClick: PropTypes.func,
   target: PropTypes.string,
-  rel: PropTypes.string
+  rel: PropTypes.string,
+  customStyle: PropTypes.shape({})
 };
 
 AnchorButton.defaultProps = {
   type: "primary",
   onClick: () => {},
   target: "_self",
-  rel: ""
+  rel: "",
+  customStyle: {}
 };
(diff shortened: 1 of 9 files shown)


src/button/Button.jsx
```diff
@@ -1,35 +1,59 @@
...
+  onClick,
+  type,
+  href,
+  target,
+  rel,
+  children
+}) {
+  const Base = props =>
+    props.as === "button" ? (
+      <button {...props}>{props.children}</button>
```
paschalidi: now that I have a second look this is kinda problematic. I will refactor

paschalidi:
image
the Anchor button can be seen here

lanwen:
anchor button seems not so aligned with text as clickable one

paschalidi:
image
image

paschalidi:
image
image

src/button/button.style.js
```diff
@@ -110,33 +141,34 @@ const ButtonStyles = styled.div`
     }
   }
 
+  .quarternary .anchor-child,
```
paschalidi: my editor says that there is a typo here quaternary not quarternary. Can someone confirm and then I will fix it

paschalidi:
image
image

paschalidi:
image
image

paschalidi:
image
image

stories/button.story.jsx
```diff
@@ -4,6 +4,8 @@ import { storiesOf } from "@storybook/react";
 import { withInfo } from "@storybook/addon-info";
 import { action } from "@storybook/addon-actions";
 import Button from "../src/button/Button";
+import AnchorButton from "../src/button/AnchorButton";
```
herman-rogers: Yea this is the right choice, having it as two different components rather than passing in a prop to change it.

src/button/AnchorButton.jsx
```diff
@@ -0,0 +1,46 @@
...
+import ButtonStyles from "./button.style";
+
+export default function AnchorButton({
+  onClick,
+  type,
+  href,
+  target,
+  rel,
+  children,
+  customStyle
```
herman-rogers: Do we need custom style? You can always just add your own styles via css/styled components. Worst case you can always add !important

paschalidi: I added for cases like where u want to override the width a default button has, it might be also that you would like to change the font-size or font-weight, stuff like that.

herman-rogers:
Locally the font is not aligned and the text is wrong:
image

paschalidi:
image
image

lanwen:
primary with image - image should go down for 1px

paschalidi:
this can be set by the Icon component by using the prop `style`. I can hardcoded but I cannot guarantee that I will look the same with every icon we will like to have next to the text

paschalidi:
image
image

lanwen:
Now secondary and quarternary should be 1px top

Also I can show you magic of the loops to check all images with buttons ;) 
Components are used everywhere, so they should be pixel perfect - as its easier to achieve that on isolated env.

Also why you copying all those images, adding a test tag to show state of the current branch will be the solution to show it in action

paschalidi:
> Also why you copying all those images, adding a test tag to show state of the current branch will be the solution to show it in action

can u explain that to me please?

paschalidi:
image
image

herman-rogers:
This is legacy - but theres also a weird button highlight when a button is active - can you remove that since we're updating them anyways.

paschalidi:
done @herman-rogers . Can you please confirm if that's a typo or not `quarternary`. Google would say its `quaternary`.

src/button/button.style.js
```diff
@@ -1,38 +1,71 @@
 import styled, { css } from "styled-components";
 
-const baseButton = css`
+const baseStyles = css`
   display: flex;
   align-items: center;
   justify-content: space-evenly;
-  font-size: 1.125em;
+  cursor: pointer;
+  padding-left: 10px;
```
herman-rogers: actually instead of hardcoding these padding values, you use use flex box to handle the layout - if you need to you can change the content justifying to a different layer then add padding to the text or icons. It's just really hard to cover all permutations with padding.

paschalidi: yeah I will go for it

paschalidi: you are right we can add padding to the children of the component itself. I was following the default button styling and that includes padding on the left and right. 
I guess for just very long text etc etc. I will remove the paddings now

paschalidi: image
Just let me know what are the requirements and I will go for it. I am just explaining why padding was there

lanwen:
>can u explain that to me please

you can setup gh-page publish based on branch or tag and spend less time getting screenshots and just giving the link

paschalidi:
>you can setup gh-page publish based on branch or tag and spend less time getting screenshots and just giving the link
I will start with that when I finish the things I have on my place 👍

herman-rogers:
Yea it's quaternary

lanwen (approved):

herman-rogers:
So there's a few things missing on these buttons:

For Anchor buttons:
- Tertiary is missing its hover state, and active state
- So is the quarternary call to action (plus the title is lower case "quaternary call to action")
- We also don't need that really long button with a huge block of text
- Secondary button is missing hover and active state
- Primary is also missing hover and active state

For Normal buttons:
- We're missing the mouse pointer on hover

paschalidi:
will go through it. Thanks

herman-rogers (approved):

PR #31, opened 2018-10-02
MAIN -  adds `stage-0` and `es2015

paschalidi:
* adds `stage-0` and `es2015` to `.babelrc`
* adds Attribute `download` to `AchorButton` component
* fixes `Button` component bug with parsing styles


.babelrc
```diff
@@ -1,3 +1,3 @@
 {
-  presets: ['env', 'react'],
-}
+  "presets": ["env", "es2015", "stage-1", "react"]
```
herman-rogers: this is stage-1 not stage-0

package.json
```diff
@@ -53,19 +61,7 @@
...
     "@storybook/storybook-deployer": "^2.3.0",
-    "babel-cli": "^6.26.0",
-    "babel-core": "^6.26.0",
-    "babel-jest": "^23.0.1",
-    "babel-loader": "^7.1.2",
-    "babel-preset-env": "^1.7.0",
-    "babel-preset-react": "^6.24.1",
-    "babel-runtime": "^6.26.0",
     "css-loader": "^0.28.9",
-    "eslint": "^4.17.0",
```
herman-rogers: Shouldn't these remain dev dependencies?

src/button/AnchorButton.jsx
```diff
@@ -2,25 +2,10 @@ import React from "react";
...
     <ButtonStyles>
-      <a
-        className={type}
-        onClick={onClick}
-        href={href}
-        target={target}
-        rel={rel}
-        style={customStyle}
-      >
+      <a {...rest} className={type}>
```
herman-rogers: what is rest?

paschalidi: rest is the rest of the props that might get parsed. 
This includes everything. `style={{color:'red'}}`, `innerRef={()=>{}}`, `rel='nonopener'` etc etc

herman-rogers:
What is this PR for exactly?

paschalidi:
Yesterday we came across the issue that on one app there was a prop that was not easy to predict that it might be used. 
Including `stage-0` solves that as it has this feature https://codeburst.io/use-es2015-object-rest-operator-to-omit-properties-38a3ecffe90 so i included it in .babelrc and refactored the components accordingly. 

On a second look removing unused packages belongs to a different PR, so I will revert all the changes regarding on that.

src/button/AnchorButton.jsx
herman-rogers: Can't we just define the rel='noopener' inside the a tag itself. I get that the ...rest passes in all props but I don't think its a good pattern to use. If we need to pass a lot of props liks this then this class should just be called custom button or something with a <Button>{children}</Button> feature. That way we're not extending it into the future to cover the many different ways to do a tags. Then people can just do <CustomButton><a href="whatever they want" /></CustomButton>. All people will really care about are the styles so it looks like a button.

stories/button.story.jsx
```diff
@@ -57,130 +56,117 @@ storiesOf("Buttons", module).add(
 
       <h2>Primary Call to Action</h2>
       <div className="list">
-        <Button onClick={() => click("button clicked!")}>Call to Action</Button>
-        <Button onClick={() => click("button clicked!")}>
-          <Icon name="call-contact" />
-          Call to Action
+        <Button>
```
paschalidi: Correct me if I am wrong here,
 I have the feeling here we just write plain html and using styled components we parse css classes on the tags.
Although it gives me the feeling this is not a component. As all the logic of the component is being repeated every time we want to create a Button or a Button that is an anchor under the hood.
The `<Button >` in this case is just css classes.
 
I believe we abstract this so much that we repeat ourselves every-time we want to create a Button. 

Is it that I get what u explained wrong here?

herman-rogers: So that's not what I explained. I said if it were me I'd still have a general use button component for the common case and a CustomButton component that uses button styles but does not enforce its children. That's option one. Option two would be to have a Button and a AnchorButton with  specified props. For example, we would enforce for the anchor button to always have the rel property set, and only allow a few other properties we specifically design. **We should not**, however, pass in a {...props} object that allows everyone to pass in whatever they want. I know that this is convenient and seems to solve the problem of catching all use cases but it has definite downsides. I can write a blog post about the downside but it seems pretty obvious to me - we're limiting our ability to change the component for the future, we're not allowing an explicit api to be set for users of the button, we're not guaranteeing types or checking for safety, etc. Plus there could even be considerations on how do React's optimizations work with dynamic props such as this (perhaps it compiles it down correctly and then can run its optimizations).

As a general rule always try to limit the complexity of a thing you want to use. It doesn't matter if its a component, function, variable, etc. Always try to guarantee a specific state that can 100% persist. This same idea applies to not using boolean toggles. That is, the various complexities and states that can occur with toggle = !toggle are greater and more unpredictable than () => toggle = false; () => toggle = true.

paschalidi: Which option do we want to go with ?

paschalidi: I will go with what u suggested.

herman-rogers: I usually go with option 2 and if absolutely necessary option 1

herman-rogers: Forgot about a third option - which is a sort of pseudo OOP, where you create a base component (say CustomButton), then create two components that extend the base, Button and AnchorButton. Although not sure if its really necessary in this case, but felt like noting it was relevant.

herman-rogers:
Why not just do ```Option two would be to have a Button and a AnchorButton with specified props. For example, we would enforce for the anchor button to always have the rel property set, and only allow a few other properties we specifically design.``` TBH I think you had the right idea first and it was fine until the {...rest} bits were added.

paschalidi:
Changed the description of the PR to 
* adds stage-0 and es2015 to .babelrc
* adds Attribute `download` to AchorButton

PR #37, opened 2018-10-16
MAIN - restructures some parts of the project 

paschalidi:


package.json
```diff
@@ -17,25 +17,20 @@
   "scripts": {
     "linter:fix": "prettier --write '{,!(node_modules|.git|storybook-static)/**/}*.{js,json,jsx,md}'",
     "linter:check": "eslint stories/ && eslint src/",
-    "storybook": "start-storybook -p 3006 -s ./public",
-    "release": "node build.js",
-    "release:test": "node build.js test",
+    "serve": "start-storybook -p 3006 -s ./public",
```
herman-rogers: make sure the infrastructure we're using is also updated with this command (shouldn't be used but double checking can't hurt).

src/Input/input.style.js
```diff
@@ -15,7 +9,6 @@ export const InputStyles = styled.div`
     font-stretch: normal;
     line-height: 2.5;
   }
-
```
herman-rogers: leave the spacing hear - otherwise its harder to reason about.

src/MedicalCard/MedicalCard.jsx
```diff
@@ -5,7 +5,7 @@ import SimpleIcon from "./SimpleIcon";
 import DetailedIcon from "./DetailedIcon";
 import missingAvatar from "../../public/images/empty-avatar.svg";
 import MedicalCardStyles from "./medicalCard.style";
-import DateDisplay from "../dates/DateDisplay";
+import { DateDisplay } from "../index";
```
herman-rogers: We shouldn't be doing this. It should be ideally ```import DateDisplay from "../dates/``` or the original form. I know the idea is to enforce using the index but we should do this inside the stories, not inside our actual source.

stories/Buttons.story.jsx
```diff
@@ -3,9 +3,7 @@ import styled from "styled-components";
 import { storiesOf } from "@storybook/react";
 import { withInfo } from "@storybook/addon-info";
 import { action } from "@storybook/addon-actions";
-import Button from "../src/button/Button";
-import AnchorButton from "../src/button/AnchorButton";
-import Icon from "../src/icons";
+import { Button, AnchorButton, Icon } from "../src/index";
```
herman-rogers: this should be enough to guarantee we're exporting our components. We should do this in source, however.

paschalidi: in source?

src/MedicalCard/MedicalCard.jsx
paschalidi: I will revert it, but it is the same thing. index.js is being bundled and served and it is already there locally.

stories/Buttons.story.jsx
herman-rogers: shouldnt* do this in source, however

src/MedicalCard/MedicalCard.jsx
herman-rogers: well in normal source we wouldn't use a concept like this, index.js only exists to serve the library / packaging tools we're using. Its not unfeasible that these tools would change and require, for example, a index.src.js file or something.

package.json
paschalidi: ✔️

stories/Buttons.story.jsx
paschalidi: ✔️

herman-rogers (approved):

herman-rogers (changes requested):
Remove all API changes

herman-rogers (approved):

PR #49, opened 2018-10-19
Design improvements

jimi-duck:
Updated text colours Primary + Secondary 
Typography aligned with design files ( padding to link for mobile)   
Input font styles, weight, hover states out, 
Button font size + weight + disabled state updated


herman-rogers:
Just an fyi - your styles will fail unless you keep in line with the eslint standards. Luckily for all of us we don't have to pay attention to the stands, you can run "npm run linter:fix" to sort if for you. 👍 to developer lazyness hah.

herman-rogers (approved):

PR #58, opened 2018-10-31
Font update

james-ciclitira:
Vivy icon font updated
All 160+ icon included
Config.json included can be used to regenerate & add using http://fontello.com/ 
Names in the existing webapp differ...


herman-rogers:
Can you update the current font story (stories/Icon.story.jsx) with the new naming? (Don't worry about adding all 160 hah we'll do that over the course of time)

james-ciclitira:
hmmm it should be done all 160ish.. haven't updated the component icons tho

PR #59, opened 2018-10-31
MAIN add forms, validation, and updated inputs

herman-rogers:
- Add improved inputs and forms to our component library
- Use the validator implementation https://github.com/yiminghe/async-validator


src/Buttons/Button.jsx
```diff
@@ -3,18 +3,21 @@ import PropTypes from "prop-types";
...
   type,
+  htmlType,
+  onClick,
   children,
+  disabled,
   customStyle
 }) {
   return (
     <ButtonStyles>
+      {/* eslint-disable-next-line */}
```
herman-rogers: https://github.com/yannickcr/eslint-plugin-react/issues/1555

Ignored the eslint error as it seems buggy imo. Would rather just have this one prop than three separate button components with the exact same content.

herman-rogers:
Should that happen in the story or in the README? Down for either tbh.

src/Buttons/Button.jsx
paschalidi: when I remove this comment I dont get an eslint error. Mb double check if we need this

src/Checkbox/checkbox.style.js
```diff
@@ -0,0 +1,59 @@
...
+    cursor: pointer;
+  }
+
+  input[type="checkbox"] {
+    display: none;
+  }
+
+  input[type="checkbox"] + label:after {
+    font-family: "vivy-icons";
+    content: "\\e80f";
```
paschalidi: isnt there a better way to do that so that we dont have to copy paster these "\\e80f" around?
mb import it in the file and use it as variable?

src/Input/Input.jsx
```diff
@@ -2,36 +2,36 @@ import React from "react";
...
+}
 
 Input.propTypes = {
   id: PropTypes.string.isRequired,
+  name: PropTypes.string,
   type: PropTypes.string,
   placeholder: PropTypes.string,
   label: PropTypes.string,
-  isRequired: PropTypes.bool
+  optional: PropTypes.bool
```
paschalidi: mb  `isOptional` so we know is boolean?
```suggestion
  isOptional: PropTypes.bool
```

src/Input/input.style.js
```diff
@@ -1,44 +1,71 @@
...
+
+    &::placeholder {
+      color: #bfbfbf;
+    }
   }
 
-  input::placeholder {
-    color: #cdcdcd;
+  .invalid {
+    border-color: #e94119;
```
paschalidi: i see we kinda using this color quite frequently here. mb add it in the `Colors` js file we have and import it from there

src/Buttons/Button.jsx
herman-rogers: So both of your comments you're not getting eslint errors - perhaps your linter doesn't pick up on some things? In the past on some of your PR's you missed some linted things so I'm skeptical of this.

paschalidi: > So both of your comments ...

which is the other commend ?

paschalidi: I mean I deleted one comment but that is irrelevant with the eslint. It was about the "\\e80f" thingy.

lanwen (approved):

src/Buttons/Button.jsx
paschalidi: > hmm weird, why did you delete that one though?

it makes sense that I add it again now that I am thinking of it. :)

src/Checkbox/checkbox.style.js
```diff
@@ -0,0 +1,60 @@
...
```
paschalidi: can import it as variable from the `public/fonts/vivy-icons/iconMapping.json` file we have?
This way if we update the vivy-icons at some point this will be updated as well and wont break if `e80f ` is not valid anymore.

paschalidi (approved):

PR #61, opened 2018-11-02
MAIN - flexgrid

paschalidi:
a very simple implementation of a flex grid.


src/FlexGrid/FlexColumn/index.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+export default function Col({ children, col, offset, style }) {
+  const classCol = col && `col-${col}`;
```
herman-rogers: just call this column

paschalidi: so is as you said that i prefix with class because the names are taken from the props.
I will leave it like it is if thats fine

src/grid/Col/index.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+export default function Col({ children, col, offset, style }) {
+  const classCol = col && `col-${col}`;
```
herman-rogers: We should probably still call it column and think of something for offset. In any case, do we need the && here since we have default props?

src/grid/Row/index.js
```diff
@@ -0,0 +1,49 @@
...
+  ]),
+  space: PropTypes.oneOf(["normal", "around", "between", "evenly"]),
+  verticalAlign: PropTypes.oneOf(["top", "middle", "bottom"]),
+  position: PropTypes.oneOf(["start", "center", "end"]),
+  reverse: PropTypes.bool
+};
+
+Row.defaultProps = {
+  style: {},
+  children: null,
```
herman-rogers: should this just be required?

src/grid/Row/index.style.js
```diff
@@ -0,0 +1,217 @@
...
+  .row {
+    box-sizing: border-box;
+    display: flex;
+    flex: 0 1 auto;
+    flex-direction: row;
+    flex-wrap: wrap;
+  }
+
+  .row.reverse {
+    flex-direction: row-reverse;
```
herman-rogers: can this not be 

`css
.row {
  .reverse {
    flex-direction: row-reverse
  }
}
`

stories/Grid.story.jsx
```diff
@@ -1,33 +1,312 @@
...
+          <Box />
+        </FlexColumn>
+      </FlexRow>
+    ))
+  )
+  .addWithJSX(
+    "offset",
+    withInfo()(() => (
+      <div>
+        <FlexRow space="around" style={{ marginBottom: 10 }}>
```
herman-rogers: Can we not repeat this style? For example, should we have a container grid class that allows us to apply styles to all the columns and rows so we don't have to repeat ourselves.

herman-rogers:
Failing build as well - missing reference.

src/grid/Row/index.style.js
paschalidi: i tried it didnt work out.

PR #66, opened 2018-11-12
MAIN - adds property onChange on Input

paschalidi:


herman-rogers:
Discussed this in IRL with @paschalidi . Basically we can have all the data pass through by the Form we're using rather than tracking individual state - so this is redundant. Closing this for now and we can reopen if we really, really, need it.

PR #74, opened 2018-11-13
MAIN - expands footer props 

paschalidi:
expands props 
1. `position: PropTypes.oneOf(["left", "center"])` for aligning on the left side when needed.
2.  `withBorder: PropTypes.bool` for adding/removing the top border when needed.


paschalidi:
no border: 
image
left: 
image

herman-rogers:
So we should create a footer that is just an empty component that fixes something to the bottom, then have children and styles passed into it (maybe in the future we need to right align it or center it). We should try to make this generic (unlike v1) so we can reuse it.

src/Footer/footer.style.js
```diff
@@ -1,53 +1,11 @@
 import styled from "styled-components";
 
 const FooterStyles = styled.div`
```
herman-rogers: Styles instead of FooterStyles

src/Footer/footer.style.js
```diff
@@ -1,53 +1,11 @@
...
-    border-top: solid 0.5px #cdcdcd;
-    margin: 0 15%;
-  }
-
-  .link {
-    display: flex;
-    border-right: 1px solid #cdcdcd;
-    padding: 0 16px;
+  a {
+    display: block;
```
herman-rogers: are you sure this doesn't do weird things?

src/VivyFooter/VivyFooter.jsx
```diff
@@ -0,0 +1,44 @@
+import React from "react";
+import Styles from "./vivyFooter.style";
```
herman-rogers: you could probably just call this file "footer.style.js" even though we have it already. We don't use it in a fashion where we'll get confused imo

paschalidi: `styles.js` ?

stories/Footer.story.jsx
```diff
@@ -1,55 +1,172 @@
...
-  "default",
-  withInfo("The raw footer component")(() => (
-    <LayoutStyles>
-      <Footer>{footerChildren}</Footer>
-    </LayoutStyles>
-  ))
-);
+storiesOf("Footer", module)
+  .add(
+    "default",
```
herman-rogers: let's not call this default, I mean, what does it even mean to be default? Is this the default that's always used (with the links inside), does it mean something else, will this always be imported, dunno. Maybe just Example default footer, Example default styles, etc.

herman-rogers: Example /w default styles - dunno this is our documentation after all - we shouldn't worry about length but more about description.

paschalidi: indeed. I renamed it to default styles

src/Footer/footer.style.js
paschalidi: margin wont work otherwise.
I tested it.

PR #77, opened 2018-11-14
MAIN - adds font family on Input and story to show how.

paschalidi:


stories/Input.story.jsx
```diff
@@ -36,7 +41,20 @@ storiesOf("Input", module)
...
+      </PageLayout>
+    ))
+  )
+  .add(
+    "icon",
+    withInfo()(() => (
+      <PageLayout>
+        <Input
+          id="password"
+          placeholder="&#xe801;"
```
herman-rogers: Does this work with both the icon and the placeholder text. Also can we use the mapping from the json file (I think you mentioned that on another PR somehweres).

paschalidi: yeah i know right?! It doesnt let me add any values here. wont get the concept. 
stupid. I will try to figure out a way around it, but I have no ideas for the time being.

> Does this work with both the icon and the placeholder text. 
yes.

herman-rogers: Placeholder={‘${Icon} password’} ?

paschalidi: its not working. But also we dont need this PR as the icon needs to be placed not as a placeholder but as an icon on the left. 
That means that even when the user types something in the field he should be able to see the icon, but when its placeholder he cant. 

So if you fine I can go ahead and close this PR.

herman-rogers: Yea that's fine go ahead and close if not needed

PR #80, opened 2018-11-14
MAIN - adds classes for icon on left

paschalidi:


src/Input/input.style.js
```diff
@@ -22,9 +23,16 @@ const InputStyles = styled.div`
...
     font-weight: normal;
     color: #353f41;
-    padding: 2px 16px;
     transition: border-color 0.2s ease, box-shadow 0.2s ease;
 
+    &.default-padding {
+      padding: 2px 16px;
+    }
+
+    &.extra-padding {
```
herman-rogers: Instead of extra why not icon-padding? That way we know what it is for.

src/Input/input.style.js
```diff
@@ -5,6 +5,7 @@ const InputStyles = styled.div`
   width: 100%;
 
   label {
+    position: relative;
```
herman-rogers: Make sure this is responsive and looks good on small and large screens

src/Input/input.style.js
```diff
@@ -22,9 +23,16 @@ const InputStyles = styled.div`
     font-size: 18px;
     font-weight: normal;
     color: #353f41;
-    padding: 2px 16px;
     transition: border-color 0.2s ease, box-shadow 0.2s ease;
 
+    &.default-padding {
```
herman-rogers: Also, why not just put this in the parent class and override the style with the icon padding inside that class

src/Input/input.style.js
herman-rogers: Also is the & required?

paschalidi: & is required yes.

src/Input/input.style.js
paschalidi: i dont get that. please elaborate. also do you see something really wrong with the current implementation ?

paschalidi: sorry miss-clicked resolve

herman-rogers: You don’t need the default class it’s redundant. This will also simplify the login in the component.

paschalidi: I thought it helps with readability. But but yeah I will go ahead

src/Input/input.style.js
paschalidi: done

src/Input/input.style.js
paschalidi: its because it is like `input.extra-padding`

src/Input/Input.jsx
```diff
@@ -25,13 +38,15 @@ Input.propTypes = {
...
 };
 
 Input.defaultProps = {
   type: "",
   name: "default",
   placeholder: "",
   label: "",
-  optional: false
+  optional: false,
+  iconName: undefined
```
herman-rogers: Make this “” instead of undefined. Empty strings are falsy. It’s just a bit more understandable.

herman-rogers (approved):

PR #82, opened 2018-11-15
MAIN - adds error classes in Inputs

paschalidi:
image
image

### DIFF
--- a/src/Input/Input.jsx
+++ b/src/Input/Input.jsx
@@ -3,9 +3,18 @@ import PropTypes from "prop-types";
 import InputStyles from "./input.style";
 import Icon from "../Icon";
 
-export default function Input(props) {
-  const { id, label, optional, type, placeholder, name, iconName } = props;
+export default function Input({
+  id,
+  label,
+  optional,
+  type,
+  placeholder,
+  name,
+  iconName,
+  errors
+}) {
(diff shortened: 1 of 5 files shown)


src/Input/Input.jsx
```diff
@@ -3,9 +3,18 @@ import PropTypes from "prop-types";
...
+  label,
+  optional,
+  type,
+  placeholder,
+  name,
+  iconName,
+  errors
+}) {
   const padding = iconName ? "icon-padding" : "";
+  const error = id in errors ? "invalid" : "";
```
lanwen: since its boolean - better to call it `invalid` or `isError`

stories/Input.story.jsx
```diff
@@ -56,4 +56,37 @@ storiesOf("Input", module)
...
+    "error",
+    withInfo()(() => (
+      <PageLayout>
+        <Input
+          id="password"
+          placeholder="Your password"
+          label="Password"
+          type="password"
+          errors={{
+            password: [{ message: "look there is an error", field: "password" }]
```
lanwen: please add example with multiple errors

paschalidi: image

lanwen (approved):

PR #83, opened 2018-11-15
MAIN - replaces button with div as form would submit when button is pressed.

paschalidi:
* runs `npm run lint:fix`
* replaces `button` with `div` as form would submit when button is pressed

### DIFF
--- a/src/InputPassword/inputPassword.style.js
+++ b/src/InputPassword/inputPassword.style.js
@@ -1,17 +1,23 @@
 import styled from "styled-components";
 
 const InputPasswordStyles = styled.div`
-  .button-position-right {
+  .icon-position-right {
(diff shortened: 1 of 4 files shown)


src/InputPassword/inputPassword.style.js
```diff
@@ -1,17 +1,23 @@
...
+    display: flex;
+    align-items: center;
     right: 0;
-    background: none;
-    color: inherit;
-    border: none;
-    padding: 0;
-    cursor: pointer;
-    outline: inherit;
+    -webkit-touch-callout: none;
```
lanwen: should it be autoprefixer job?

lanwen (approved):

PR #85, opened 2018-11-16
MAIN - adds responsive Grid for Tablets 

paschalidi:
I dont import Screenshots as this requires you to change the width of your screen

### DIFF
--- a/src/Grid/Col/index.js
+++ b/src/Grid/Col/index.js
@@ -25,8 +25,8 @@ Col.propTypes = {
 Col.defaultProps = {
   style: {},
   offset: 0,
-  xs: 12,
-  sm: 12,
-  md: 12,
-  lg: 12
+  xs: 0,
+  sm: 0,
+  md: 0,
+  lg: 0
 };
--- a/src/Grid/Col/index.js
+++ b/src/Grid/Col/index.js
@@ -1,9 +1,12 @@
 import React from "react";
 import PropTypes from "prop-types";
 
-export default function Col({ children, col, offset, style }) {
+export default function Col({ children, xs, sm, md, lg, offset, style }) {
   return (
-    <div style={style} className={`col-${col} col-offset-${offset}`}>
+    <div
+      style={style}
+      className={`col-xs-${xs} col-sm-${sm} col-md-${md} col-lg-${lg} col-offset-${offset}`}
+    >
       {children}
     </div>
   );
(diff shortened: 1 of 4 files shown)


herman-rogers:
What is the issue and what is this PR for?

lanwen (approved):

paschalidi:
I will try to make sense of this PR by explain the use case. 

there are the case that we need to have responsive design and our current implementation wont work for smaller devices. 
So I expanded it. To better understand I will try to explain a story I wrote. 
```
<Row position="center">
        <Col col={3} colMd={7}>
          <Box />
        </Col>
        <Col col={3} colMd={8}>
          <BoxBlue />
        </Col>
        <Col col={3} colMd={10}>
          <Box />
        </Col>
</Row>
```
There is a row with width equal to `3` and the same column has width equal to `7` for smaller devices. 

I will add these screenshots and tell me if it makes sense.
Larger Screens: image
Smaller Screens: 
image

paschalidi:
@herman-rogers i wont merge till you accept as this is not blocking me. I just prepare this for next week that I have to implement some designs for smaller screens.

paschalidi:
So if i get that correctly (most likely i don't tho) the idea is that we automate the responsiveness in a way that we dont have to specify the col width for Tablets or Mobile but it would be handled by css automatically? 

Like for example when we have a row with 3 columns one next to the other on tablet would be a row with 3 columns one under each other? 
Correct?

src/Grid/Col/index.js
```diff
@@ -12,12 +15,18 @@ export default function Col({ children, col, offset, style }) {
 Col.propTypes = {
   children: PropTypes.node.isRequired,
   style: PropTypes.shape({}),
-  col: PropTypes.number,
-  offset: PropTypes.number
+  offset: PropTypes.number,
+  xs: PropTypes.oneOf([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, "none"]),
```
herman-rogers: is none apart of the spec for PropTypes.oneOf?

paschalidi: none you can use when you wanna hide a column in one of the screens. So for example in you want in Tablet to hide a Column or something then you should use `sm="none"`

stories/Grid.story.jsx
```diff
@@ -37,6 +42,38 @@ storiesOf("Grid", module)
...
+    ))
+  )
+  .addWithJSX(
+    "responsive with none",
+    withInfo()(() => (
+      <Row position="center">
+        <Col xs={7} sm={12} md={7} lg={3}>
+          <Box />
+        </Col>
+        <Col xs={7} sm="none" md={10} lg={3}>
```
herman-rogers: Would rather just leave out sm as a property to denote “none” rather than have “none”

src/Grid/Col/index.js
herman-rogers: Why not “hide”?

stories/Grid.story.jsx
herman-rogers: Actually if "none" means "to hide" then I think we should think about renaming this. Otherwise I assume it means I don't want anything set for "sm" mode, rather than hide the column when small

paschalidi: The default prop is `0`. zero means the column takes the width of the content inside it.
Will add a lot of boilerplate code if we set default to `none`.

src/Grid/Col/index.js
paschalidi:
```
  .col-lg-none{
    display: none;
  }
```

this is the class that would make the display none. I thought it is close naming to the native css classes so thats why. 
If it still doesnt make sense I dont mind renaming it

herman-rogers (approved):

PR #98, opened 2018-11-22
MAIN - Extends Input

paschalidi:
* adds a wrapper for Input an InputPassword as these two share a lot in common.
* replaces InputPassword `div` onClick={} with a `button` for accessibility reasons
* expands story to toggle password visibility

### DIFF
--- a/src/InputPassword/inputPassword.style.js
+++ b/src/InputPassword/inputPassword.style.js
@@ -1,18 +1,20 @@
 import styled from "styled-components";
 
 const InputPasswordStyles = styled.div`
-  .icon-position-right {
-    font-size: 30px;
-    cursor: pointer;
+  position: relative;
+  width: 100%;
+
+  .button-position-right {
     width: 54px;
     height: 54px;
-    display: flex;
-    justify-content: center;
     position: absolute;
-    display: flex;
-    align-items: center;
     right: 0;
-    user-select: none;
+    background: none;
+    color: inherit;
+    border: none;
+    padding: 0;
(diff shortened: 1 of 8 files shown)


src/InputPassword/inputPassword.style.js
```diff
@@ -1,18 +1,20 @@
 import styled from "styled-components";
 
 const InputPasswordStyles = styled.div`
-  .icon-position-right {
-    font-size: 30px;
-    cursor: pointer;
+  position: relative;
```
lanwen: why not flex based?

paschalidi: because the button that we add on the right side of the Input needs to be there positioned `absolute`. 
Wanna see some screenshots?

lanwen: that will be cool

paschalidi: image
image

this icon on the right side is being positioned there by the class `.button-position-right`, and for absolute to work nicely I made the parent component `position: relative;`

src/InputSheath/inputSheath.style.js
```diff
@@ -2,10 +2,10 @@ import styled from "styled-components";
 import { error, errorBackground } from "../Colors";
 
 const InputStyles = styled.div`
+  position: relative;
```
paschalidi: @lanwen the previous `position: relative;` is obsolete because of the existence of this one.

lanwen (approved):

PR #112, opened 2018-11-28
MAIN - render props form validator

paschalidi:
* renames InputSheath to InputParent 
* adds Validator component with the RenderProps pattern to validate forms 
* validates fields now onChange if prop onChange is parsed on the Input component
* validates fields now onBlur if prop onBlur is parsed on the Input component
* shows errors.


stories/Forms.story.jsx
```diff
@@ -162,4 +72,70 @@ storiesOf("Forms", module)
...
-  .add("form with validation", withInfo()(() => <FormWithValidation />));
+  .add(
+    "form with validation",
+    withInfo()(() => (
+      <Validate rules={rules}>
+        {({
+          onSubmit,
+          onBlur,
+          onChange,
+          errors: { name, lastName, email, terms, newsletter }
```
herman-rogers: what does this do?

stories/Forms.story.jsx
```diff
@@ -162,4 +72,70 @@ storiesOf("Forms", module)
...
+        }) => (
+          <Row position="center">
+            <Col col={9}>
+              <Form submit={onSubmit}>
+                <Input
+                  onBlur={onBlur}
```
herman-rogers: what happens if you don't pass in onBlur?

stories/Forms.story.jsx
paschalidi: its the same like if I would do const  `{ name, lastName, email, terms, newsletter } = errors` 3 lines later.

stories/Forms.story.jsx
paschalidi: wont validate onBlur

paschalidi:
@herman-rogers morning, after some research I came to the conclusion that using (Context API)[https://reactjs.org/docs/context.html] of react is a very good way to share functionality around the components. (I could also do it with hooks but this API is still under proposal). Please review the changes I made here and let me know your thoughts.

src/Forms/Validation.jsx
```diff
@@ -0,0 +1,83 @@
...
+    }
+  };
+
+  validateForm = e => {
+    if (e && e.preventDefault) {
+      e.preventDefault();
+    }
+
+    const { rules } = this.props;
+    const fields = Array.from(e.target.elements)
```
herman-rogers: is e.target.elements already an array?

src/Forms/Validation.jsx
```diff
@@ -0,0 +1,83 @@
...
+const DEFAULT_STATE = {
+  errors: {}
+};
+
+export const ValidationContext = createContext(DEFAULT_STATE);
+
+export default class Validation extends Component {
+  state = DEFAULT_STATE;
+
+  valueOfTarget = target => {
```
herman-rogers: why not just call this "fieldValue" or "inputValue" or something. We should avoid using keywords in variable names. What are keywords? Things like bool / target / class / export / const / etc.

paschalidi: sure

src/Forms/Validation.jsx
paschalidi: no its not already an array. why do you ask?

herman-rogers: Thought it was an array

paschalidi: ehm, tbh this piece of code was there. I will double check.

src/MedicalCard/MedicalCard.jsx
```diff
@@ -27,7 +27,8 @@ function MedicalCard({ user, data, t }) {
               <DateDisplay
                 date={pregnancyDueDate.data}
                 monthNames={t("component.months")}
-              />)
+              />
+              )
```
herman-rogers: is this prettier?

paschalidi: u run linter:fix

PR #113, opened 2018-11-29
MAIN - hotfix

paschalidi:


herman-rogers:
Footer.story.jsx seems to have a col={0} property.

src/Grid/Col/index.js
```diff
@@ -1,11 +1,31 @@
...
+  children,
+  lg,
+  md = lg,
+  sm = md || lg,
+  xs = sm || md || lg,
+  offsetLg,
+  offsetMd = offsetLg,
+  offsetSm = offsetLg || offsetMd,
+  offsetXs = offsetLg || offsetMd || offsetSm,
+  style
```
paschalidi: this looks strange. I know. 
What would this do is 
*  md = lg ===  set the md = lg when md is not parsed as property
*  sm = md || lg === to set the sm = md when sm is not parsed and to lg when not even md is parsed 
etc etc

src/Grid/Row/index.style.js
```diff
@@ -16,21 +16,16 @@ const RowStyles = styled.div`
...
       flex-direction: column;
     }
   }
 
-  .col {
-    .reverse {
-      flex-direction: column-reverse;
-    }
-  }
-
```
paschalidi: this is unused therefore I deleted it

src/Grid/Col/index.js
```diff
@@ -15,18 +35,24 @@ export default function Col({ children, xs, sm, md, lg, offset, style }) {
 Col.propTypes = {
   children: PropTypes.node.isRequired,
   style: PropTypes.shape({}),
-  offset: PropTypes.number,
-  xs: PropTypes.oneOf([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, "hide"]),
-  sm: PropTypes.oneOf([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, "hide"]),
+  offsetXs: PropTypes.oneOf([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]),
+  offsetSm: PropTypes.oneOf([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]),
+  offsetMd: PropTypes.oneOf([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]),
+  offsetLg: PropTypes.oneOf([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]),
```
paschalidi: I added offset for all the devices as well. So that we can define different offsets for different devices when needed

src/Grid/Col/index.js
herman-rogers: Erm...not sure what that means? Haha

paschalidi: haha I will try to rephrase.
So the point is that when someone is having a column 
```
<Col lg=10> some things here </Col>
```
then in my understanding is that, it would be nice when `md=10` and `sm=10` and `xs=10` without one having to explicitly define all of those every time a Col is being defined.
A value should fallback to the previous defined value.

So this way when someone has define `lg=10` (for instance) all the other properties will fallback to the value `10`

Does how this makes more sense ?

herman-rogers: I see, is there a more intuitive way we can do this in the code? Maybe inside the render itself - I get what you're going for but I think a refactor is in order to make the idea more clear.

PR #115, opened 2018-12-03
MAIN - extends grid

paschalidi:
adds   
```
  positionMd,
  positionSm,
  positionXs,
  textAlign,
  textAlignMd,
  textAlignSm,
  textAlignXs,
```
to handle position of the flex box and text alignment for different devices.


herman-rogers:
Can you explain in detail what you’re trying to achieve?

PR #116, opened 2018-12-04
MAIN - adds form action

paschalidi:
* adds form action
* adds submit button on form without any validation
* makes rule of name an array

### DIFF
--- a/src/Forms/Form.jsx
+++ b/src/Forms/Form.jsx
@@ -7,9 +7,7 @@ export default function Form({ children, select, action }) {
     if (e && e.preventDefault) {
       e.preventDefault();
     }
-    if (typeof validation === "undefined") {
-      submitAction();
-    } else if (validation(e) === null) {
+    if (!validation || validation(e) === null) {
       submitAction();
     }
   };
--- a/src/Forms/Form.jsx
+++ b/src/Forms/Form.jsx
@@ -2,13 +2,22 @@ import React from "react";
 import PropTypes from "prop-types";
 import { ValidationContext } from "./Validation";
 
-export default function Form({ children, select }) {
+export default function Form({ children, select, action }) {
+  const submit = (e, validation, submitAction) => {
+    if (e && e.preventDefault) {
+      e.preventDefault();
+    }
+
+    if (validation(e)) {
+      submitAction();
+    }
+  };
   return (
     <ValidationContext.Consumer>
-      {({ onSubmit }) => (
+      {({ validation = () => true }) => (
         <form
           onChange={e => select(e.target.value)}
-          onSubmit={e => onSubmit(e)}
+          onSubmit={e => submit(e, validation, action)}
         >
           {children}
         </form>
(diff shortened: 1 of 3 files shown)


src/Forms/Form.jsx
```diff
@@ -2,13 +2,22 @@ import React from "react";
...
 import { ValidationContext } from "./Validation";
 
-export default function Form({ children, select }) {
+export default function Form({ children, select, action }) {
+  const submit = (e, validation, submitAction) => {
+    if (e && e.preventDefault) {
+      e.preventDefault();
+    }
+
+    if (validation(e)) {
```
herman-rogers: what happens if validation is undefined?

src/Forms/Validation.jsx
```diff
@@ -39,9 +36,11 @@ export default class Validation extends Component {
           (prev, error) => ({ ...prev, [error.field]: error.message }),
           {}
         );
+        isValid = false;
       }
       this.setState({ errors: err });
     });
+    return isValid;
```
herman-rogers: so instead of doing a boolean does the result from the validator package not return false? If not, we could also just return the errors and use that to check if there are errors or not.

stories/Forms.story.jsx
```diff
@@ -88,7 +95,7 @@ storiesOf("Forms", module)
     withInfo("Using the Form without any validation")(() => (
       <Row position="center">
         <Col lg={9}>
-          <Form>
+          <Form action={() => console.log("an rxjs action")}>
```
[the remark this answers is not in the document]
paschalidi: ah yes

stories/Forms.story.jsx
```diff
@@ -111,7 +121,7 @@ storiesOf("Forms", module)
       <Row position="center">
         <Col lg={9}>
           <Validation rules={rules}>
-            <Form>
+            <Form action={() => console.log("an rxjs action")}>
```
herman-rogers: Same here

src/Forms/Form.jsx
```diff
@@ -2,13 +2,22 @@ import React from "react";
...
+      submitAction();
+    }
+  };
   return (
     <ValidationContext.Consumer>
-      {({ onSubmit }) => (
+      {({ validation = () => true }) => (
```
paschalidi: @herman-rogers  when validation comes in undefined here it is gives a default value.

src/Forms/Form.jsx
paschalidi: https://jenkins.uvitatech.com/job/UvitaTeam/job/uvita-webapp/job/master/290/console
made a comment there

herman-rogers: in jenkins?

src/Forms/Form.jsx
herman-rogers: so the default value is true?

src/Forms/Validation.jsx
paschalidi: yes we can return the errors and when there are no errors then `validator` package return null so this will work indeed.

src/Forms/Form.jsx
paschalidi: is a function that returns true.

herman-rogers: we should just allow it to pass undefined and check if its undefined. Less state yea?

paschalidi: and when it is undefined I should trigger the action?

herman-rogers: when its undefined we assume there's no validation thus only trigger the submit

PR #117, opened 2018-12-04
MAIN - fixes broken styles

paschalidi:
* cleans footer from deprecated class and story  
* fixes colors story 
* makes full-width story from Grid more understandable
* removes redundant Input stories


stories/Footer.story.jsx
```diff
@@ -131,42 +131,4 @@ storiesOf("Footer", module)
         </Container>
       </Footer>
     ))
-  )
```
paschalidi: all of the stories have no border, to add border you need to explicitly add a css class

PR #118, opened 2018-12-05
MAIN- fixes classnames

paschalidi:
* fixes the bug with several spaces on html classes


stories/Grid.story.jsx
```diff
@@ -377,13 +377,11 @@ storiesOf("Grid", module)
     ))
```
paschalidi: please dont let this bit get you confused.
I just reordered the stories so that the responsive story goes after the text-align-right story.

utils/classNames.js
```diff
@@ -0,0 +1,9 @@
+export default function(...args) {
+  return args.reduce((prev, index) => {
+    const [className, condition = true] = index;
```
herman-rogers: what does condition mean?

paschalidi: so i considered the case that 
```
  const row = classNames([`row`]);
```
that means the condition (above) is `undefined`. When is `undefined` initial value is true.

I think the moment you know what this does actually makes the code cleaner. 
There are ofc many ways to do that so if you wanna go with some other way please let me know

herman-rogers: so, if the className doesn't exist, then add it? Seems strange?

src/Grid/Row/index.js
```diff
@@ -81,7 +73,7 @@ Row.defaultProps = {
   style: {},
   space: "normal",
   verticalAlign: "top",
-  position: "start",
+  position: "sindextart",
```
herman-rogers: typo?

paschalidi: yes

herman-rogers:
Double check that the NPM build command will pull in the source code from the utils folder since it lies outside of the src/ folder. It may not package that code together.

PR #131, opened 2018-12-07
MAIN - footer

paschalidi:
makes footer border top from 100 percent to max-width 960px


herman-rogers:
does this also remain responsive on smaller screens?

PR #144, opened 2019-01-03
MAIN - updates stories and messages.

paschalidi:
Will update just the messages of the grid story and the styles of the boxes. 
Has no actual css changes that will affect any of the components.

### DIFF
--- a/stories/Grid.story.jsx
+++ b/stories/Grid.story.jsx
@@ -28,7 +28,7 @@ storiesOf("Grid", module)
   .addWithJSX(
     "default width",
     withInfo(
-      "would have the default maximum width of the column which is 960px"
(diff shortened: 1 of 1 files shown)


stories/Grid.story.jsx
```diff
@@ -4,47 +4,42 @@ import { withInfo } from "@storybook/addon-info";
...
+  flex: 1;
 `;
 
 storiesOf("Grid", module)
   .addWithJSX(
-    "default",
-    withInfo()(() => (
+    "default width",
+    withInfo(
+      "would have the default maximum width of the column which is 960px"
```
herman-rogers: Use complete sentences please. "The Row would have the default max....960px." or "The Grid" or whatever subject you're referring to.

PR #145, opened 2019-01-04
MAIN Add default value to input

herman-rogers:
- Add default value for inputs
- Fix errors with missing validation context

### DIFF
--- a/stories/Input.story.jsx
+++ b/stories/Input.story.jsx
@@ -2,7 +2,7 @@ import React from "react";
 import styled from "styled-components";
 import { storiesOf } from "@storybook/react";
 import { withInfo } from "@storybook/addon-info";
-import { Input, InputPassword } from "../src/index";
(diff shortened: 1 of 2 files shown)


stories/Input.story.jsx
```diff
@@ -23,14 +23,14 @@ storiesOf("Input", module)
     ))
   )
   .add(
-    "optional",
+    "default value",
```
lanwen: I think better to add also optional example without default value

herman-rogers: optional doesn't do anything afaik. It isn't even a prop on the input

herman-rogers: And the "basic" story, shows an input without any default values.

lanwen: anyway, better to have example with absent default value

lanwen (approved):

PR #158, opened 2019-01-14
MAIN Add message notification components

herman-rogers:
image


src/index.js
```diff
@@ -1,3 +1,6 @@
+import "react-toastify/dist/ReactToastify.min.css";
```
paschalidi: does this really need to be here?

herman-rogers: as in? Put it elsewhere?

src/Message/Message.js
```diff
@@ -0,0 +1,70 @@
...
+      progressClassName: "progress-bar"
+    });
+  },
+  error(text, options = {}) {
+    return toast.error(
+      <MessageContent
+        icon="failed-filled-16"
+        text={text}
+        color="#e94119"
+        size={18}
```
paschalidi: why would this be size?

paschalidi: can we also rename this to fontSize?

src/Message/Message.js
```diff
@@ -0,0 +1,70 @@
...
+  color: PropTypes.string,
+  size: PropTypes.number
+};
+
+MessageContent.defaultProps = {
+  color: "#04d4bf",
+  size: 22
+};
+
+export function MessageContainer() {
```
paschalidi: why not `export default`?

paschalidi:
Do you mind if we name this Notification(s)?
`Message` is too generic and we might need that in the future for some sort of "Message". 
I believe Notifications is more what this popup thingy is.

src/Message/Message.js
```diff
@@ -0,0 +1,70 @@
...
+  return (
+    <Styles>
+      <ToastContainer />
+    </Styles>
+  );
+}
+
+export const message = {
+  ...toast,
+  info(text, options = {}) {
```
paschalidi: can you explain, maybe with a story how this `options` can be used?

src/index.js
paschalidi: yes maybe we can have it in the file `src/Message/Message.js`? i have the feeling it is more part of the component than part of the exports.

src/Message/Message.js
herman-rogers: it's too big otherwise

paschalidi: Cant all then be 18px?

src/Message/Message.js
herman-rogers: because there's multiple exports inside of here.

src/Message/Message.js
herman-rogers: There's a documentation story that explains it

src/Message/Message.js
herman-rogers: So they're not all uniform in size. If they are all the same size some of them will be larger than others. The Error icon is particularly bigger than the rest.

paschalidi: I see. Thanks

paschalidi (approved):

PR #160, opened 2019-01-15
Update notify text styles

herman-rogers:
Fix some small issues with the notification text styles


src/Notify/notify.style.js
```diff
@@ -11,6 +11,11 @@ const Styles = styled.div`
   .text-content {
     display: flex;
     align-items: center;
+
+    p {
+      text-align: left;
```
paschalidi: isnt that the default?

paschalidi (approved):

PR #161, opened 2019-01-16
MAIN - updates styled components

paschalidi:


.storybook/customDecorators.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import Fonts from "../src/Fonts/fonts.style";
+import { withOptions } from "@storybook/addon-options";
+
+export function withGlobalStyles(storyFn) {
+  return (
+    <React.Fragment>
```
herman-rogers: import {Fragment} from "react" -> Fragment

paschalidi:
Closed this PR as storybook version `4.1.6` had a minor bug that would throw an error when a `story.jsx` has  any styled-component included on the file 
for instance 
```
const PageLayout = styled.div`
  padding: 50px 100px;
  display: flex;
`;
```
.

This  lead me to a witch hunting which was unnecessary and ugly but through this I have figured some things out that I wanna keep for later on the road map, therefore I will keep this branch locally and delete it from the `origin`.
I am opening a new PR that will address the exact same thing as this one.

PR #164, opened 2019-01-17
MAIN - update styled components

paschalidi:
* updates styled-components to v4
* exports Fonts.
* separates storyshots into multiple instead of one monolithic file.


package.json
```diff
@@ -24,14 +24,6 @@
     "test:watch": "npm run test -- --watchAll"
   },
   "dependencies": {
-    "@storybook/addon-actions": "^4.1.6",
```
paschalidi: lets talk about this change. I think when we minify and distribute this UI package having all these dependencies here is not necessary.  
When we import this UI package into a project as `doctors-service` we dont really need any storybook there.
I think this way the bundle will be smaller therefore the loadings faster. 
Let me know your thoughts @herman-rogers, i might be wrong.

herman-rogers: happy to optimize the builds so that we only have the lib and no story book in them.

package.json
```diff
@@ -16,22 +16,14 @@
   ],
   "scripts": {
     "linter:fix": "prettier --write '{,!(node_modules|.git|storybook-static)/**/}*.{js,json,jsx,md}'",
-    "linter:check": "eslint stories/ && eslint src/",
+  "linter:check": "eslint src/**/* && eslint stories/*.jsx",
```
herman-rogers: missing a space here?

PR #183, opened 2019-01-26
MAIN - adds danger in the CI

paschalidi:
This is a PR that test out Danger.js and the value it could bring to our projects.


paschalidi:
Testing out Danger.js and the value it could bring to our projects.

paschalidi:
@lanwen sorry for the spam.

lanwen (approved):

paschalidi:
@lanwen i have paused these because we work on encryption. but thanks for approving

PR #184, opened 2019-01-29
MAIN masked input

paschalidi:
screencapture-localhost-3006-2019-01-29-14_22_32


stories/Forms.story.jsx
```diff
@@ -156,6 +160,15 @@ storiesOf("Forms", module)
                 label="Last name"
                 optional
               />
+              <InputMask
+                mask="1111 1111 1111 1111"
+                id="card"
+                placeholder="Please add your card number"
```
lanwen: better to use our examples, since its our library for medical services

lanwen:
don't really like that this lib will be bundled for every service which wants to use components. How can we avoid that?

paschalidi:
when you say lib you mean the `react-maskedinput`? In general the best approach for a library like this would be that we have the flexibility to add each component separately if we like to.
For example have only `@vivy-components/forms` installed and not the whole package. 

This can be done with lerna easily. 
https://github.com/TerraEclipse/react-stack
look in `modules` file for reference.

This way each component will be released separately that means if for example in the Future that if `Forms` have a breaking change, and a service doesnt want this change it can avoid it by just not installing it. Then it can only install some other packages. 

Le me know if that makes sense

paschalidi:
well another solution is to implement out own masked Input inspired by the `react-maskedinput`.

stories/Forms.story.jsx
```diff
@@ -21,6 +22,9 @@ const rules = {
       message: "Minimum characters is 8"
     }
   ],
+  card: [
+    { type: "string", required: true, message: "Card number is required" }
```
lanwen: wrong error :)

lanwen (approved):

PR #226, opened 2019-02-13
Add alert components

Weetbix:
:wave: This PR: 
- Adds 2 new Alert components
- Updates the icon line-height to always be 1. Previously when using an icon inside a page where the line-height is styled, the icon would also use this, which leads to incorrect dimensions for the icon.


Weetbix:
Our repo isn't setup correctly for Danger to run on PRs from forks. See this issue: https://github.com/danger/danger-js/issues/675

As per the danger docs we should create an open-source token with only permission to write comments to PRs, then set this token to be output into the build log - although it's a bit weird to expose the token like that.

For now I'll close this in favor of a branch on upstream: https://github.com/UvitaTeam/vivy-components/pull/230

PR #258, opened 2019-03-07
Add Table component

Weetbix:
This PR adds a new Table component to the Vivy-Component library

Leverages React-Table which is popular, hugely customizable, supports sorting and pagination with both clientside and serverside data sources and a bunch more stuff. This will allow our table to be extensible in the future to support these features. 

I've removed the default conflicting styles of the table and added our own styles, and some sensible defaults. We wrap the React-Table element which allows all the features to pass through and be usable (pending styling...).

Check out the screenshots in this PR for how it looks 👍 

@paschalidi one for you 🙂


paschalidi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>Big PR - <i>This PR is unlikely to get reviewed because it touches too many lines 
  (5877). 
  Consider sending smaller Pull Requests and stack them on top of each other.</i></td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 923c4bf8dce1d949410fddaa3b37f80e207ae75e
</p>

.eslintrc.json
```diff
@@ -4,6 +4,7 @@
   "plugins": ["prettier"],
   "rules": {
     "import/no-extraneous-dependencies": false,
+    "import/prefer-default-export": false,
```
paschalidi: 👍 if you also agree, we should do that in every project actually.

package.json
```diff
@@ -36,6 +36,7 @@
     "react-dom": "^16.8.4",
     "react-i18next": "^7.12.0",
     "react-input-mask": "2.0.4",
+    "react-table": "^6.9.2",
```
paschalidi: I might would argue that since version 7 is out we should go for it.. I know its is alpha but lets at least have this on the table for discussion.

src/Table/Table.jsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import { StyledTable } from "./table.styles";
+
+const tableWithDefaults = props => {
```
paschalidi: I would go an explicitly define the arguments that are coming as props.
This way we can have Type checking using React `prop-types`.

src/Table/Table.jsx
```diff
@@ -0,0 +1,16 @@
...
+  const defaultOptions = {
+    showPagination: false,
+    minRows: 0,
+    sortable: false,
+    resizable: false
+  };
+
+  return <StyledTable {...defaultOptions} {...props} />;
+};
+tableWithDefaults.propTypes = StyledTable.propTypes;
```
paschalidi: `StyledTable.propTypes;` having hard time understand what is this value. Can you please explain :)

src/Table/table.styles.js
```diff
@@ -0,0 +1,71 @@
+import ReactTable from "react-table";
+import styled from "styled-components";
+import "react-table/react-table.css";
+import colors from "../Colors";
+
+const WithoutConflictingDefaultStyles = styled(ReactTable)`
```
paschalidi: cant `WithoutConflictingDefaultStyles` and `StyledTable` be merged into one const? Would that make sense to you? 
Just for me reading the code dont see the reasoning behind splitting it.

stories/Table.story.jsx
```diff
@@ -0,0 +1,122 @@
...
+    ];
+
+    return (
+      <LayoutStyles>
+        <Table data={medicalPersonsData} columns={columns} />
+      </LayoutStyles>
+    );
+  })
+  .add("custom cell contents", () => {
+    // eslint-disable-next-line react/prop-types
```
paschalidi: i think i might just define those prop-types. dont have a strong opinion tho.

.eslintrc.json
Weetbix: Yeah I agree :)

I suggest we create an eslint rules package like `eslint-config-airbnb` to keep these things easy and consistent

paschalidi: yeah. good idea

package.json
Weetbix: What are the arguments for using the alpha version?

In general we should use the latest stable version of depdendencies. v7 is alpha and under heavy development so I would strongly recommend we don't use it. We are likely to run into many bugs as it's a complete re-write, and 6 is nice and stable. For v7 there are not even docs in the Readme

src/Table/Table.jsx
Weetbix: Here we just use the existing `ReactTable` proptypes. As we are just wrapping `ReactTable` and setting some sensible defaults, we allow all proptypes from `ReactTable`. If we extend this later to blacklist or add some more propTypes we can do stuff like this:
```
tableWithDefaults.propTypes = {
    ...StyledTable.propTypes, 
    extraProp: PropTypes.string.isRequired
};
```

src/Table/Table.jsx
Weetbix: We have typechecking from prop-types (see below comment)

src/Table/table.styles.js
Weetbix: Ah ok. I split them so there is a clear distinction between which styles we are stripping from react-table, and which ones we are adding as extras. 

If that still doesn't make sense, I can remove it?

stories/Table.story.jsx
Weetbix: I opted not to because it's just an anonymous component I'm using immediately, proptypes seems unnecessary

package.json
paschalidi: Okay read through their issues. Apparently v7 will be out soonish.. v6 tho is not actively maintained anymore.. Hope we dont run into blockers with this :)

src/Table/table.styles.js
paschalidi: Makes sense. It is best this way now that you explained. Maybe add comment?

stories/Table.story.jsx
paschalidi: yeah okay.

src/Table/table.styles.js
Weetbix: Good idea :)

src/Table/Table.jsx
paschalidi: Now i see.

PR #301, opened 2019-03-27
Do not use empty strings for download or rel attributes

Weetbix:
Currently the anchor buttons defaults for the `download` and `rel` prop is an empty string. 

This means that all `AnchorButtons` will have the `download` attribute set, and instead of navigating will download the page. 

This PR changes the default to null, which means the `download` attribute wont be set on the anchor.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,9 +10,9 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-          - v1-dependencies-{{ checksum "package.json" }}
-          # fallback to using the latest cache if no exact match is found
-          - v1-dependencies-
+            - v1-dependencies-{{ checksum "package.json" }}
+            # fallback to using the latest cache if no exact match is found
+            - v1-dependencies-
 
       - run: npm install
       - run: npm run build:static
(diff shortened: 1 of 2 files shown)


paschalidi:
@Weetbix makes sense. please check why test fail. I would try to create again all the images and commit. For some reason this is needed sometimes. It has to do with updating jest. I will approve then.

paschalidi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
      <td>✅ Get a 🥇 for opening a small enough PR. - <i>Be proud of yourself!</i></td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against efbf586110925fdda10e7a1bb6029c0cd7495f28
</p>

Weetbix:
@paschalidi as discussed, updating locally would not produce new images for some reason, and re-running on circleci fixed the issue this time

For the future, I've updated the circle ci config to store the diffs as a build artifact, so that we can see them in the circle ci UI next time it happens:

Issue #201, opened 2019-02-06
Create components for HeaderOne HeaderTwo etc etc 


paschalidi:
We need to align with design first (I will arrange a meeting) and then we should implement these components so that we use them around our apps.

PR #323, opened 2019-04-02
MAIN - implements components Display, H1, H2

paschalidi:
👋 👋 Very good day!! 

Closes https://github.com/VivyTeam/vivy-components/issues/201

#### Introduction
This PR will add three components. `Display`, `HeadingOne`, `HeadingTwo`. I didnt decide the naming instead I took it from the design. (you can find Zeplin designs later on this description)

#### Few clarifications
You might wondering why `Display` component is a `div` and not `h1`.  The answer is that the design team would like to use this for slogans and still dont have an `h1` tag on it. 
Why you ask? 
Because of sso, the idea is that `HeadingOne` would have a more descriptive text that would used for sso and `Display` would be some slogan here and there.

#### Extras
There are images as snapshots where you can see the components and compare it.
Here you can also fine the link of the Zeplin to quickly check with the original design.
1. Zeplin app zpl://screen?sid=5bf57e4b5544f67ec5077aa7&pid=5a8d80ddc1a5d9014b9a9b29
2. Zeplin on web zpl.io/2vN3gMP


paschalidi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 0f2837f0b0814ece8b6a99632e2630ef18fbe046
</p>

src/Display/Display.jsx
```diff
@@ -0,0 +1,21 @@
...
+  font-weight: 500;
+  font-size: 48px;
+`;
+
+export default function Display({ children, style }) {
+  return <DisplayStyles style={style}>{children}</DisplayStyles>;
+}
+
+Display.propTypes = {
+  children: PropTypes.string.isRequired,
```
Weetbix: The HTML standard says the childen of a header tag can be any phrasing content: https://html.spec.whatwg.org/multipage/dom.html#phrasing-content

So I think the children proptypes for all of these should be something like:
```
PropTypes.oneOfType([
    PropTypes.arrayOf(PropTypes.node),
    PropTypes.node
  ]),
```

So it children can be other nodes.

src/Display/Display.jsx
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import PropTypes from "prop-types";
+import styled from "styled-components";
+
+const DisplayStyles = styled.div`
```
Weetbix: I like your explaination about what Display is and why it's not a header element in the PR :) Lets put that as living documentation in the story, or as a comment here?

stories/H2.story.jsx
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import { H2, Colors } from "../src/index";
+
+storiesOf("HeadingTwo", module)
```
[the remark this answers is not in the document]
paschalidi: > I also wonder if all these typographies belong better under one "typography" story header, but we can do that in a PR later if we both agree

I dont know, lets decide. I can also add it there but since its components maybe it needs also to be their own stories?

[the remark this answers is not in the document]
Weetbix: Yeah, I mean like under a `Typography` subtree or something, I think we can do it by just adding `Typography/` to the start of the story names https://storybook.js.org/docs/basics/writing-stories/#nesting-stories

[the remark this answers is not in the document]
paschalidi: I am wondering if we want the structure of the storybook to be different that the structure of the `src` in our project.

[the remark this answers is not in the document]
Weetbix: Yeah true, just worried that it will be messy if they arent grouped together, lets wait and see :)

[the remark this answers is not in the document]
paschalidi: i dont mind if we group them together or not. But maybe we want to group them inside the `src`? 
anyway lets wait and see. no big deal to change this around 👍

src/Display/Display.jsx
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import PropTypes from "prop-types";
+import styled from "styled-components";
+
+// This is a div element that is in font-size bigger than h1 and is supposed to
```
Weetbix:
```suggestion
// This is a div element that is in font-size bigger than h1 and is supposed to be
```

src/Display/Display.jsx
```diff
@@ -0,0 +1,28 @@
...
+// used as 'slogan' around our apps. We dont want this to be crawled from search engines
+// since its only for design reasons but instead we want the h1 to be so as it is meant by default.
```
Weetbix:
```suggestion
// since its only for design reasons.
```
I think we can drop the end of this (or rewrite it?)

src/Display/Display.jsx
```diff
@@ -0,0 +1,28 @@
...
+const DisplayStyles = styled.div`
+  font-weight: 500;
+  line-height: 1.13;
```
Weetbix: Where did you get this lineheight value from? seems should be either 1.16 or 56px?

src/H1/H1.jsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import PropTypes from "prop-types";
+import styled from "styled-components";
+
+const Styles = styled.h1`
+  font-weight: 500;
+  line-height: 1.17;
```
Weetbix: Same for this one? should be either 1.16 or 42px?

src/H2/H2.jsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import PropTypes from "prop-types";
+import styled from "styled-components";
+
+const Styles = styled.h2`
+  font-weight: 500;
+  line-height: 1.25;
```
Weetbix: this one seems to be correct

src/Display/Display.jsx
paschalidi: ehm when you click on the element from Zeplin then you can see the styling of the component. This is where I take the styles from always.

Weetbix: Ahhhhhh, I see. Weird its different to what the actual text says right next to it. Can you check with piotr which one is correct?

paschalidi: rem sounds the most correct.

Weetbix: Can you explain why? Also does that mean we should be using rem for font size?

paschalidi: i dont have a strong reason I think out of the top of my mind that rem is more responsive. 
I will convert to pixel you are right.

Weetbix: Take some time and look into it maybe we should be using that here instead, not against it just need some reasoning

paschalidi: uhuh will do then.

PR #336, opened 2019-04-10
MAIN - adds modal

paschalidi:
👋 hello 🇦🇺 

#### Motivation 
We are gonna need a `Modal` that is controlled from its parent rather than being rendered each time. This PR tries to improve the api of the `Modal` that already exists. 

#### Why is `body-scroll-lock` istalled?
Prevent scrolling when `Modal` is open can tricky to implement for all the browsers.  After [reading around](https://medium.com/jsdownunder/locking-body-scroll-for-all-devices-22def9615177) I decided to go for [a package](https://www.npmjs.com/package/body-scroll-lock) that provides an out of the box solution. Feel free to disagree on this. [Are you worried of the the bundle size?](https://bundlephobia.com/result?p=body-scroll-lock@2.6.1)

#### Why the width of this modal is not dynamic? 
I will make a pr that will introduce different widths later on when we know what is the actual usecases for this component. I cant really decide now what the widths should be. If you wanna propose smth feel free.

#### Why accessibility has not been added?
Thought of adding it in a new PR since its kinda of a "feature"

#### Why Colors has been moved to the top?
Because Modal styles cant access it if its being imported before the actual `Modal` import.

#### Why in this story there are `hooks`?
Its only an example of a component that might exist in other apps. No functionality of our core components includes any `hooks`. Dont hesitate to disagree

Looking forward to your feeback!! 🐶


paschalidi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 5554fc0bc96f32d8993ed5d270853da099897bac
</p>

src/Modal/Modal.jsx
```diff
@@ -1,20 +1,37 @@
...
-      ) : null}
-    </ModalStyles>
-  );
+class Modal extends Component {
+  componentDidMount() {
+    disableBodyScroll(document.querySelector("body"));
+  }
+
+  componentWillUnmount() {
+    clearAllBodyScrollLocks();
```
Weetbix: Better to call `enableBodyScroll` on the same element that you called disable on here (`body`) rather than disable all scroll locks as you might inadvertently disable someone elses

src/Modal/modal.style.js
```diff
@@ -1,32 +1,37 @@
...
-    left: 50%;
+const zIndex = 10;
+export const Overlay = styled.div`
+  position: absolute;
+  left: 0;
+  top: 0;
+  width: 100%;
+  height: 100%;
+  z-index: ${zIndex};
+  background-color: rgba(0, 0, 0, 0.7);
```
Weetbix: Where did 0.7 come from? seems that our concepts use 0.3?

src/Modal/Modal.jsx
```diff
@@ -1,20 +1,37 @@
...
 }
 
 Modal.propTypes = {
-  children: PropTypes.node.isRequired,
-  open: PropTypes.bool.isRequired
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.node),
+    PropTypes.node
+  ]).isRequired,
+  closeCallback: PropTypes.func
```
Weetbix: Lets be consistent with our prop names, this should be named `onClose` (see button, anchor etc)

src/Modal/modal.style.js
```diff
@@ -1,32 +1,37 @@
...
+  left: 0;
+  top: 0;
+  width: 100%;
+  height: 100%;
+  z-index: ${zIndex};
 `;
 
-export default ModalStyles;
+export const Content = styled.div`
+  z-index: ${zIndex};
```
Weetbix: Where are you getting these stylings from? They dont match our new concept or the old modal?

Our new concept has no box shadow, bigger border radius etc?

paschalidi:
@Weetbix when built that 2 days ago there was no concept. And thought will add the style on a different PR. 
Now since concept is there I will add the correct styling on this PR

paschalidi:
@Weetbix please have another look when your time allows it 👍

PR #337, opened 2019-04-12
MAIN - Removes outliner in an accessible way

paschalidi:
♿️ since we want to be health care cool startup who care about all the users, I thought of improving some accessibility on our `a`nchors and `button`s.

#### Motivations
How things are at the moment is that there is no `outline` on the button when you navigate using tabs.
I think that is not what out product should be like. 
I have found a solution that will show the `outline` only when user is navigating with tabs and not when actually clicking the button. Little bit as everybody else does it. 

To rephrase, the final result will be that the outliner is there when you navigate through your tabs and will not be there when you click the button unless the focus wont go anywhere else and stay in the button. 

If the above is not super clear I would be happy to explain in person. :) 

Please have a look and let me know. 🌈

### DIFF
--- a/src/Buttons/button.style.js
+++ b/src/Buttons/button.style.js
@@ -32,8 +32,11 @@ const ButtonStyles = styled.div`
     border-radius: 4px;
     display: block;
     text-decoration: none;
-    outline: none;
     cursor: pointer;
+  }
+
+  button:focus:active {
+    outline: 0;
   }
 
   a {
--- a/src/Buttons/button.style.js
+++ b/src/Buttons/button.style.js
@@ -35,16 +35,17 @@ const ButtonStyles = styled.div`
     cursor: pointer;
   }
 
-  button:focus:active {
-    outline: 0;
-  }
-
   a {
     font-family: system-ui;
     border-radius: 4px;
     display: inline-block;
     text-align: center;
     text-decoration: none;
+  }
+
+  a:focus:active,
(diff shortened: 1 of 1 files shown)


paschalidi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against fe62e3ac7500c54759bad2f1350bf1255b30f154
</p>

lanwen:
please don't use emojis in commit messages and pr titles - you can commit them accidentally, but that's sometimes not well supported in a text-based git log viewers

paschalidi:
sure

lanwen:
Just find something like video recorder - https://www.cockos.com/licecap/

paschalidi:
to do what with it? I prefer if we implement PR preview for CircleCI than doing this manually for each PR I make here.

lanwen:
I mean you can use licecap to demo the difference

paschalidi:
anchroButton
Button

paschalidi:
With this change when one presses tab it will be able to see the outliner, when before was impossible to actually know in which button you are.

lanwen (approved):

PR #349, opened 2019-04-16
MAIN - adds optional buttons

paschalidi:
Hey you amazing reviewers 🌻 

I am adding the buttons in this version of the Modal.

**Styling changes** 
Also adds some styling changes that needed to be done.
1. Adds `Height ` this helps aligning the modal vertically.
2. Adds max-width on modal of 920px as in concept.

🕊 hope you enjoy reviewing this one! peace out!

**Who said  g i f ?!** 🤠 😎 
Modal-gif


paschalidi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 5d2630077eee639704dececdb470d9198aa70659
</p>

lanwen:
Shouldn't ok/cancel btns have different colours from the beginning?

src/Modal/Modal.jsx
```diff
@@ -40,16 +41,40 @@ class Modal extends Component {
...
+                          type="tertiary"
+                          onClick={onClose}
+                          style={{ marginRight: 16 }}
+                        >
+                          Cancel
+                        </Button>
+                      </Col>
+                      <Col lg={0}>
+                        <Button type="tertiary" onClick={onSubmit}>
+                          Submit
```
lanwen: should be translatable

src/Modal/Modal.jsx
```diff
@@ -40,16 +41,40 @@ class Modal extends Component {
...
+                  <BodyContent>
+                    {children}
+                    <Row position="end">
+                      <Col lg={0}>
+                        <Button
+                          type="tertiary"
+                          onClick={onClose}
+                          style={{ marginRight: 16 }}
+                        >
+                          Cancel
```
lanwen: should be translatable

src/Modal/Modal.jsx
```diff
@@ -40,16 +41,40 @@ class Modal extends Component {
...
+                      </CloseButton>
+                    </Col>
+                  </Row>
+                  <BodyContent>
+                    {children}
+                    <Row position="end">
+                      <Col lg={0}>
+                        <Button
+                          type="tertiary"
+                          onClick={onClose}
```
lanwen: should be `onCancel` and btn should be there only if callback is defined

src/Modal/Modal.jsx
```diff
@@ -40,16 +41,40 @@ class Modal extends Component {
...
+                          style={{ marginRight: 16 }}
+                        >
+                          Cancel
+                        </Button>
+                      </Col>
+                      <Col lg={0}>
+                        <Button type="tertiary" onClick={onSubmit}>
```
lanwen: btn should be there only if callback is defined. To allow no-btn modal

stories/Modal.story.jsx
```diff
@@ -27,7 +29,10 @@ storiesOf("Modal", module)
             Launch composed modal
           </Button>
           {open && (
-            <Modal onClose={() => toggleModal(false)}>
+            <Modal
```
lanwen: please add all modal combinations

paschalidi: I will add when there is usecase for this. The api is obviously extendable here. 
If you both agree that this is part of this PR even though there is no use-case in our product for that to exist then I will go ahead and add it. But for now I would prefer if we dont add it.
@Weetbix @lanwen your thoughts please?

lanwen: thats quite easy to add and better to do it now, to not switch context next time. Also I think we will get usecase quite soon. In terms of why to add usecases here - since the stories is not only about usecases, but also about testing possible variants.

paschalidi: yeah sorry I deleted my comment as I started the implementation already. 
here is the comment that I deleted:
I will add when there is usecase for this. The api is obviously extendable here. 
If you both agree that this is part of this PR even though there is no use-case in our product for that to exist then I will go ahead and add it. But for now I would prefer if we dont add it.
@Weetbix @lanwen your thoughts please?

paschalidi:
@lanwen would be great if you can have another look here!!

src/Modal/Modal.jsx
```diff
@@ -29,7 +31,30 @@ class Modal extends Component {
...
+    if (onCancel && !cancelText) {
+      // eslint-disable-line
+      console.warn(
+        "You might need to add property 'cancelText' if you need the button responsible for cancel to have text"
+      );
+    }
+    if (onSubmit && !submitText) {
+      // eslint-disable-line
+      console.warn(
+        "You might need to add property 'submitText' if you need the button responsible for submit to have text"
```
lanwen: please remove the warning - its a total valid case to not have the buttons

paschalidi: the warning is not to inform you that you might not have button. this is 100 percent valid feature.
When one uses the component like 
```
             <Modal
              onClose={() => toggleModal(false)}
              onCancel={() => alert("I can cancel something")}
              onSubmit={() => alert("I can submit something")}
            >
              A very basic modal that wont scroll the background when opened.
            </Modal>
```
it is very likely that the user will forget to add 
```
  submitText="Submit"
  cancelText="Cancel"
```
properties that are for translated text. I would not like to go to storybook and I would expect a well made library to tell me what I should add to have the text.

src/Modal/Modal.jsx
```diff
@@ -62,11 +119,19 @@ Modal.propTypes = {
...
+  submitText: PropTypes.string,
   role: PropTypes.string,
   ariaLabel: PropTypes.string
 };
 Modal.defaultProps = {
-  onClose: () => {},
+  onClose: null,
+  onCancel: null,
+  onSubmit: null,
+  cancelText: "",
```
lanwen: cancel and submit can have default text

paschalidi: then this text wont be translatable. we dont translate from vivy-components

src/Modal/Modal.jsx
lanwen: well made lib will have well made defaults

paschalidi: the library dont translate... we need to parse translatable variables from our applications

src/Modal/Modal.jsx
```diff
@@ -40,16 +65,48 @@ class Modal extends Component {
...
+                              {cancelText}
+                            </Button>
+                          </Col>
+                        )}
+                        {onSubmit && (
+                          <Col lg={0}>
+                            <Col lg={0}>
+                              <Button
+                                style={{ marginLeft: 16 }}
+                                type="grey"
```
lanwen: I think gray usually means "disabled" thus suggest to be green

paschalidi: well i know. I will talk with Piotr again.. agree

src/Buttons/Button.jsx
```diff
@@ -30,7 +30,14 @@ export default function Button({
...
   children: PropTypes.node.isRequired,
   onClick: PropTypes.func,
-  type: PropTypes.string,
+  type: PropTypes.oneOf([
+    "primary",
+    "secondary",
+    "tertiary",
+    "quaternary",
+    "white",
+    "grey"
```
Weetbix: I think there might be some confusion here due to the current designs, lets discuss in person as I don't think there was meant to be a grey button, just a disabled state for the previous buttons

Also I dislike that we now have button types that are colours and a button types that are functional mixed, it doesn't make sense.

In fact I would say we rather just style the quaternary button to fit the "white" button use case as it's not use anywhere from what I can see, and then we can keep the number of different button styles down to 4 instead of 6

src/Modal/Modal.jsx
```diff
@@ -29,7 +31,30 @@ class Modal extends Component {
...
+      onCancel,
+      onSubmit,
+      role,
+      ariaLabel,
+      children,
+      submitText,
+      cancelText
+    } = this.props;
+
+    if (onCancel && !cancelText) {
```
Weetbix: This kind of stuff should be done as a custom prop type: https://stackoverflow.com/questions/42299335/react-props-set-isrequired-on-a-prop-if-another-prop-is-null-empty

src/Modal/Modal.jsx
```diff
@@ -40,16 +65,48 @@ class Modal extends Component {
...
+                            <Button type="white" onClick={onCancel}>
+                              {cancelText}
+                            </Button>
+                          </Col>
+                        )}
+                        {onSubmit && (
+                          <Col lg={0}>
+                            <Col lg={0}>
+                              <Button
+                                style={{ marginLeft: 16 }}
```
Weetbix: Should all styles belong in the styles file?

stories/Modal.story.jsx
```diff
@@ -1,22 +1,109 @@
...
+              </Modal>
+            )}
+          </>
+        );
+      };
+      return <ModalUseCase />;
+    },
+    {
+      info:
+        "Sumbit button is show since property 'onSubmit' is being passed on the component."
```
Weetbix:
```suggestion
        "Sumbit button is shown since property 'onSubmit' is being passed to the component."
```

stories/Modal.story.jsx
```diff
@@ -1,22 +1,109 @@
...
+              </Modal>
+            )}
+          </>
+        );
+      };
+      return <ModalUseCase />;
+    },
+    {
+      info:
+        "Cancel button is show since property 'onCancel' is being passed on the component."
```
Weetbix:
```suggestion
        "Cancel button is shown since property 'onCancel' is being passed to the component."
```

Weetbix (changes requested):

src/Modal/Modal.jsx
paschalidi: sometimes i get lazy.. is this really very bad here?

src/Buttons/Button.jsx
paschalidi: agree.

Weetbix:
As discussed in person we will discuss with our designer and hopefully condense down to two buttons, primary and secondary as we already have new component concepts that only have these two types

paschalidi:
Yeah will open a new PR with the Buttons only and we should focus on that one before we start reviewing this one again. Thanks!

src/Modal/Modal.jsx
lanwen: yes, it should be the responsibility of service to add translations, but still btns should be usable from the very beginning

paschalidi: yeah agree, why dont we make it that when the user has passed the prop onSubmit but not any text the propValidation would complain that a text should be added, also an error will be shown on the console? This way we restrict the user to always have text passed in the button when the usecase is to have Sumbit button shown.

paschalidi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3c6d7980537edd25477eeab5943488d75e6f6c98
</p>

paschalidi:
@Weetbix @lanwen please have another look.
Now the new buttons we created the other day are being shown. 
Also you can PR-review. 

For the future as talked with John in private there needs to be a boolean property that will have the modal button `disabled` for the time it is being false and `enabled` for the other case. 
This will help us prevent the doctors to submit the modal without first uploading documents. 

I will implement this in a different PR as it is kinda of a different "feature"

stories/Modal.story.jsx
```diff
@@ -17,6 +19,91 @@ const LayoutStyles = styled.div`
...
+    }
+  )
+  .add("both Cancel and Submit buttons", () => {
+    const ModalUseCase = () => {
+      const [open, toggleModal] = useState(true);
```
lanwen: whats the difference with the next story?

paschalidi: the modal is not shown and you click the button to trigger it..

lanwen: doesn't think that necessary, since trigger btn is already here and you can click it

paschalidi: @lanwen my bad. the last story would show that the modal wont actually scroll when you have it opened. Therefore I will add better description in the story so that everybody can tell what it is about but I think it makes sense not to mix the stories but keep it separate. 
What you think?

lanwen (approved):

stories/Modal.story.jsx
Weetbix: Not sure how I missed this the first time but:
```suggestion
        "Submit button is show since property 'onSubmit' is being passed on the component."
```

Weetbix: @paschalidi this was marked as resolved without being changed?

paschalidi: interesting. I missed that sorry.

src/Modal/Modal.jsx
```diff
@@ -62,11 +101,41 @@ Modal.propTypes = {
...
   role: PropTypes.string,
   ariaLabel: PropTypes.string
 };
 Modal.defaultProps = {
-  onClose: () => {},
+  onClose: null,
+  onCancel: null,
+  onSubmit: null,
+  cancelText: "",
+  submitText: "",
```
Weetbix: These text props shouldn't have default blank string values anymore right? As they are basically 'required' when onsubmit is passed

paschalidi: what do you think it should be?

Weetbix: no default? we are throwing a prop type warning when its not set

paschalidi: eslint complains since it is not always required but only sometimes.. any ideas?

Weetbix: Ohhh it doesnt work well with custom prop validators I guess then... can we change them to null so it's a bit more obvious they are not usable as strings?

paschalidi: When making it `null`  and because we have a custom validation that would check if the user adds as property a `string` an error will be thrown that the specific prop needs to be string and not null.

Why dont we keep it empty as it is since its an empty string ?

Weetbix: ok lol, no big deal. 

Im beginning to dislike the 'optional props' props should have default rule

Weetbix (approved):

Weetbix:
Thanks for the changes!! 🎉

src/Modal/Modal.jsx
paschalidi: yes i remember you told me that. websync

PR #386, opened 2019-05-28
MAIN - Add onChange and onBlur handlers to inputs

Weetbix:
👋 

Hello. 

Currently it's not possible to get the value from one of our inputs and process it as a consumer as there are no onBlur/onChange props exposed. 

This PR: 
- Adds onBlur and onChange properties to all inputs
- Validation Context Improvements:
  - this context used to expose `validation`, `onBlur`, `onChange` to all consumers, but they were all just calls to validation. Then consumers would write stuff like: `onBlur={ validateOnBlur ? onBlur(e) ...`. I've collapsed all these into a `validate` function it exposes instead as it's much more descriptive. 
  - updated the default values for the context. Previously every time you would type in any input not in a validation context (ie a form) you would get console errors as the onChanged/blur handlers were not defined. You can check this in the live storybook. I've added defaults to this so we don't care if they are defined or not.


paschalidi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against ed02578777325401469e1ba43c22fb29dc7458b4
</p>

src/Input/Input.jsx
```diff
@@ -30,16 +32,20 @@ export default function Input({
...
-                : e => cleanField(e.target.id)
-            }
-            onBlur={
-              validateOnBlur
-                ? e => onBlur(formData(e.target.form || {}), e.target.id)
-                : null
-            }
+            onChange={e => {
+              if (validateOnChange) {
+                validate(formData(e.target.form || {}), e.target.id);
```
paschalidi: hey jo! 
`e => onChange(formData(e.target.form || {}), e.target.id)`
this will return a function but currently there is no function being returned. How do we ensure that only 
`validate(formData(e.target.form || {}), e.target.id);`, for instance, gets triggered and not both this and `onChange(e);`
?

Weetbix: >this will return a function but currently there is no function being returned. How do we ensure that only

what do you mean? event handlers should not return a function

We _do_ call both validate and onChange, I can imagine scenarios where you want to know the value but also want the validation to still function?

paschalidi: i see

paschalidi:
hm did you have the time to  add a similar story that will explain how to use `onBlur` property?

Weetbix:
@paschalidi I added it 👍

paschalidi:
Thanks!

PR #387, opened 2019-05-31
MAIN - extends modal api

paschalidi:
Heya @Weetbix! hope you enjoy the day! 🏄‍♀
 

This PR extends the existing api of the modal by passing some new properties to it. 

* `buttonsPosition` to allow users position the buttons of the footer at the middle or the start of the `row`
* `modalLength` to allow users to change the length of the modal. 

I am attaching the Zeplin project that lead to this changes.
zpl://screen?pid=5a8d80ddc1a5d9014b9a9b29&sid=5cc31d814a92851016f011ce
zpl.io/aNm54kn
You will see a success modal on the right end of the flow.


paschalidi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against c46480aa98a0f711e9a64283f7e2b4a9f55fa82f
</p>

src/Modal/Modal.jsx
```diff
@@ -68,7 +70,7 @@ class Modal extends Component {
           <Overlay onClick={onClose} />
           <Height>
             <Row position="center" verticalAlign="middle">
-              <Col>
+              <Col lg={modalLength}>
```
Weetbix: this should be the width property then right? 

I wonder if it's smarter to just have 2 different size of modals and let the consumer choose either large or small. 
Small would have the buttons centered and be this wide, large would be what we currently have. 

The advantage of this way would be that we retain full control over how the modal would be shown in different screen widths (ie there should be no difference between large/small for xs screen sizes). If we let the users pass these values in then they will have trouble when they want to change modal size based on screen size, and will need to pass even more props.

paschalidi: makes sense but if design later on has another length how will we proceed?

Weetbix: we can discuss at that point if it makes sense to have a third width, and even then we would have less than the 7 we would have if we allow the user to pass it in. Same reasoning why we don't allow various button sizes or colours to be configured. We also need to be aware of the creep of multiple components and try to ensure that designs use the existing components where possible, or commit to a global change - otherwise we will have a very fragmented UI library and a very fragmented user experience across our apps

src/Modal/Modal.jsx
```diff
@@ -145,7 +146,8 @@ Modal.propTypes = {
   },
   role: PropTypes.string,
   submissionEnabled: PropTypes.bool,
-  ariaLabel: PropTypes.string
+  ariaLabel: PropTypes.string,
+  modalSize: PropTypes.oneOf(["sm", "lg"])
```
Weetbix: no need to prefix this property with `modal` imo. Otherwise looks good 👍

PR #391, opened 2019-06-03
MAIN - adds Select component

paschalidi:
Heya! 🌳 

What is this PR about.
I am extending the api of `vivy-components` by adding the extra component `Select` from the library https://react-select.com/

* I thought of keeping it a little bit flexible since the library has lots of features and therefore I added `...rest` as `prop`, we can talk about it in case you think it's not great solution.
* I removed a yarn.lock file that was there for some reason we use npm for this if I am not mistaken


paschalidi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 787dfb4d2a76a8eeae39f2179b8fabe5b440dcf3
</p>

src/Select/select.style.js
```diff
@@ -0,0 +1,57 @@
+import styled from "styled-components";
+import { Colors } from "../index";
+
+const OverrideSelectStyles = styled.div`
```
Weetbix: As `react-select` supports/uses the `className` prop, why dont we just dervice from the select component here, that way we are not wrapping it with an extra dom element

ie. `const blah = styled(ReactSelect)`

Weetbix: This "Override select styles" seems to be a mix of things that need to be unset or removed, and a mix of our own styling. 

Please remove only include unsets or "removals" (ie margin: 0) in this style part

paschalidi: > Please remove only include unsets or "removals" (ie margin: 0) in this style part


so we would like to have a const for removing all the style and a const for adding our own style ?

PR #401, opened 2019-06-14
MAIN - Do not show text behind the eye

Weetbix:
👋 

G'day!

This PR fixes two issues:
- Text going behind the eye icon in the password input
- Dashlane icon overlapping the eye icon.

It turns out that the dashlane eye is always attached to the `input` element itself, so this means when we want to offset the content with the button, we should instead draw the input border on the above element. So this PR changes the input wrapper so that the parent element (containing everything like the input, eye icon etc) is the one with the border (instead of the input) and then makes the input width the correct logical size (shorter). 

@paschalidi 🎈 

The dashlane icon is still a little bit overlapped by the text, but I dont think we should change our padding size in the design to compensate for some 3rd party browser plug.

Pics:
image

### DIFF
--- a/src/InputPassword/inputPassword.style.js
+++ b/src/InputPassword/inputPassword.style.js
@@ -24,6 +24,10 @@ const InputPasswordStyles = styled.div`
       font-size: 24px;
     }
   }
+
+  input {
+    margin-right: 54px;
+  }
 `;
 
 export default InputPasswordStyles;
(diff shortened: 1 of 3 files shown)


paschalidi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against c5ffc024c210d05467b15dc3bc6e93fbf0248c32
</p>

PR #425, opened 2019-07-10
MAIN - adds new colors

paschalidi:
Hey 👋 

here we have the brand new `Color` naming!  🎉 

#### More Details

* Updates all the tests
* Also whenever we import `Colors` I get it imported from the `index.js`. The reason for this is that when we import all the Components from the `index.js` we eliminate the possibility of us forgetting to export one new component. Now for the Colors that might not be the case but then this falls back into the consistency reasoning. 

Waiting for your thoughts!


paschalidi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against beea4766fbb17febeca68a667da3385063c09f02
</p>

src/Colors/index.js
```diff
@@ -3,55 +3,23 @@ export const hexToRGBA = (hex, alpha = 1) => {
...
+  affirmativeDim: "#E8F5E5",
+  warningDim: "#FFF6D1",
   textPrimary: "#353F41",
   textSecondary: "#7B7B7B",
   textInactive: "#BFBFBF",
+  textActive: "#00BFAC",
   divider: "#E0E0E0",
-
+  background: "#F8F7F5",
   backgroundPrimary: "#F5F5F5",
```
Weetbix: background primary is not in our colour pallet?

stories/Welcome.jsx
```diff
@@ -10,5 +10,5 @@ storiesOf("Welcome", module)
   .add("Principles", () => <PrinciplesPage />)
   .add("Vision & Ethos", () => <EthosPage />)
   .add("Typography", () => <ThemeTypography />)
-  .add("Colors", () => <ThemeColors />)
+  .add("Color Palette", () => <ThemeColors />)
```
Weetbix: 👍

Weetbix (changes requested):
Please go through all the image diffs in this PR, there are some big changes that should not happen. Some text turns from blue to black, modal dialog is smaller etc

src/Colors/index.js
paschalidi: you request to remove backgroundPrimary and backgroundSecondary?

Weetbix: Im asking what these are? if they arent in our colour pallet then why are they here and where did they get defined from? where are they used and what is the impact of not exporting these

paschalidi: there is not in the palette and since thats a breaking change it can be removed. For some reason I had the feeling that this were being used but I was wrong. 
Will remove and request review.

Weetbix: yes we should use this opportunity to completely align the colour exports with the design library, as we have breaking changes anyway

paschalidi:
all the colors that is not in the palette are removed and all the old references are removed. Please have another look @Weetbix

paschalidi:
@Weetbix this will now have only the guideline colors, also removes every instance of the deprecated colors from the lib.

src/Buttons/button.style.js
```diff
@@ -89,7 +89,7 @@ const ButtonStyles = styled.div`
 
     &:hover {
       ${transitionBoxShadow};
-      background-color: ${Colors.brandAlternative};
+      background-color: ${Colors.affirmative};
```
Weetbix: This is a totally different colour? is it a mistake?

Weetbix: Yes, all our buttons are green on hover now

image

Weetbix (changes requested):
Please go through and check the colours are not different from their current usage (alternative vs affirmative)

paschalidi:
@Weetbix went thought the colors and double checked. Please have another look when you get the time!

Weetbix (approved):
👍 Thanks!

PR #435, opened 2019-07-17
MAIN - fixes modal height for mobile devices

paschalidi:
Heya @Weetbix,

### About this PR

* Removes `Height` since it's not needed there.
* Also makes the Body of the modal scrollable and gives it the ability to stretch depending on the content. 
* The whole Modal has 80vh maximum height for sm devices, and 70vh for the rest as we discussed.
In the end I didnt use the `Div100vh` thingy since you cannot really add classes to it and we need to have media queries.  Plus that lib is another external dependency we can in the end avoid.

I have it tested for iphone 5, iphone 6 and a larger screen on adroid, you will find the screenshot bellow. 

## iPhone 5
IMG_0414

## iPhone 6
IMG_0161

## Android
Screenshot_20190717-172825


paschalidi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 18c5273129d0da7efe58adf8164f11bd1596d825
</p>

src/Modal/Modal.jsx
```diff
@@ -82,39 +82,43 @@ class Modal extends Component {
                       </CloseButton>
                     </Col>
                   </Row>
+                </HeaderArea>
 
-                  <BodyArea id="modal-body-area">{children}</BodyArea>
+                <BodyArea>
+                  <div id="modal-body-area" className="overflow-container">
```
Weetbix: Why do we need 2 layers here? (body area + sub div). 

You should just be able to remove the diplay: flex from Body Area and keep flex: 1 there

src/Modal/Modal.jsx
```diff
@@ -67,10 +67,10 @@ class Modal extends Component {
...
         <ModalStyles>
           <Overlay onClick={onClose} />
-          <Height>
-            <Row position="center" verticalAlign="middle">
-              <Col lg={size === "sm" ? 7 : 12} sm={12}>
-                <Content>
+          <Row position="center" verticalAlign="middle">
+            <Col lg={size === "sm" ? 7 : 12} sm={12}>
+              <Content>
+                <HeaderArea>
```
Weetbix: also not convinced that we need to add this?

paschalidi: yeah you are right in the beginning I didnt add it as well and in the browser it looks great, when the tests run though the modal doesnt look as it supposed to be looking and adding `flex: 1 0` will give the right space to the header and the bottom. 

wanna talk about it in person maybe we find a better solution?

PR #475, opened 2019-09-06
Wip

kaylalarson1990:
@cmwhited this is what I have so far. It's not dynamic at all and I want to figure out the best way to do this as the rest of the test I will be making will be exactly the same as this so I just want to be sure I'm doing this properly.


lanwen:
@kaylalarson1990 Hey, thanks a lot for your contribution! Is that list of the tests is hand-written or somehow generated?

PR #621, opened 2020-05-18
MOB 101 support typescript and convert Button to typescript

a7madgamal:
- also made styled-components peerDependency
this is a pre-step for importing the button component in the portal (MOB-101)


src/Buttons/Button.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { FC, ReactElement } from "react";
```
a7madgamal, line 1: @Weetbix would be nice to test the new version with other repos using the button just to make sure it still works

a7madgamal:
wrong ticket number 🤦

PR #631, opened 2020-06-05
MOB-244 convert Input component to TS

a7madgamal:
**Heads up 1:** Had to increase the timeout for tests. The problem is with the `quite` phase of generating snapshots (no text changed in the terminal for a very long time). I tried hard to make it more verbose or update some related packages hoping for a faster execution but failed 🤷 
**Heads up 2:** I removed what seemed to be useless `optional` prop which TS rightfully complained about in stories


src/InputWrapper/InputWrapper.tsx
```diff
@@ -20,7 +27,7 @@ export default function InputWrapper({
         {label}
         {optional ? <span className="optional">*optional</span> : null}
       </span>
-      {/* eslint-disable-next-line */}
+      {/* eslint-disable-next-line jsx-a11y/label-has-for */}
```
Weetbix, line 30: 😎 very nice

.circleci/config.yml
```diff
@@ -36,7 +36,10 @@ jobs:
           at: ~/repo
 
       - run: npm run linter:check
-      - run: npm test
+      - run:
+          name: Tests
+          no_output_timeout: 15m # a workaround until we make it faster later
```
Weetbix, line 41: Ohhhh I see now, so it was a circle ci timeout. How weird. Happy to not spend any more time on this as we will move to GH actions later

PR #638, opened 2020-06-11
v2.2.1

a7madgamal:
I already published 2.2.0 yesterday

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@vivy/vivy-components",
-  "version": "2.1.0",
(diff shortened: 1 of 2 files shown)


Weetbix:
Wait I'm confused, did you publish manually? 

We should always use a release tag and let CI handle publishing and pushing the new version. This also means we are missing the github release for this version. 

I would not merge this PR but instead create a 2.2.2 release tag and let CI do it

a7madgamal:
oh sorry I got confused with the eslint repo which doesn't have automated process yet.

PR #649, opened 2020-06-16
[EBS - 128 ] - WEB | Convert Sidebar component to Typescript

hamzahsn:
**Description:**

This PR is about converting Sidebar component to typescript

**AC:**

**out of scope:**

**Jira Link:**

https://uvitadev.atlassian.net/browse/EBS-128

**Screenshot:**


src/Sidebar/Sidebar.tsx
```diff
@@ -8,12 +7,43 @@ import {
...
+interface HeaderFC<T> extends React.FC<T> {
+  displayName: string;
+}
+
+interface SideBarFC<Link, Header, Sidebar> extends React.FC<Sidebar> {
+  Link: LinkFC<Link>;
+  Header: HeaderFC<Header>;
+}
+
+const Link: LinkFC<LinkProps> = ({
```
a7madgamal, line 41: I think it's better to separate those into different files
@Weetbix wdyt?

src/Sidebar/Sidebar.tsx
```diff
@@ -8,12 +7,43 @@ import {
...
+type SideBarProps = {
+  children: ReactNode;
+};
+
+interface LinkFC<T> extends React.FC<T> {
+  displayName: string;
+}
 
-const Link = ({ iconName, active, href, children }) => {
+interface HeaderFC<T> extends React.FC<T> {
```
a7madgamal, line 32: no need fot this extraction let's use `FC` directly in the component

src/Sidebar/Sidebar.tsx
Weetbix: I think it's alright here, better for now to avoid any unnecessary refactoring to keep PRs smaller and changes easier to review etc

src/Sidebar/Sidebar.tsx
a7madgamal: optional comment

hamzahsn: I just made a generic type to give types to `displayName`, `Header`, `Link` and its nothing fancy actually, but how do you think to avoid these ?

src/Sidebar/Sidebar.tsx
```diff
@@ -8,12 +7,43 @@ import {
...
 
-const renderableChildrenPropType = PropTypes.oneOfType([
-  PropTypes.arrayOf(PropTypes.node),
-  PropTypes.node,
-]);
+type LinkProps = {
+  iconName: string;
+  active?: boolean;
+  href: string;
+  children?: ReactNode;
```
a7madgamal, line 14: not needed. please always remove from types unless needed

src/Sidebar/Sidebar.tsx
```diff
@@ -8,12 +7,43 @@ import {
...
+};
+
+type HeaderProps = {
+  logoUrl?: string;
+  logoAltText?: string;
+  href?: string;
+  children?: ReactNode;
```
a7madgamal, line 21: same here

src/Sidebar/Sidebar.tsx
hamzahsn: hahhah, just now removed it before you comment and will push this one looool

src/Sidebar/Sidebar.tsx
hamzahsn: same here mate 😆 I just removed it

src/Sidebar/Sidebar.tsx
```diff
@@ -8,12 +7,43 @@ import {
...
+};
+
+type SideBarProps = {
```
a7madgamal, line 24: here also, SideBarProps isn't needed in that case

hamzahsn: I removed all what you said already before even you comment 😄

hamzahsn: you can take a look while it finish build and test 😉

definitions.d.ts
```diff
@@ -0,0 +1 @@
+declare module "*.svg";
```
a7madgamal, line 1: this will fix TS errors, but does svgs work correctly in storybook?

hamzahsn: yes, it works correctly in storybook and I have tested it

hamzahsn: and if we need to declare modules within our `TS/TSX` file then we can include this file into `tsconfig.types.json` file

a7madgamal: approved but not sure about the filename, @Weetbix do you have better suggestions?

hamzahsn: IMO, this is a common pattern for the name of this file as it gives a meaning for what it have also you can find others who name it: `typings.d.ts` or `declarations.d.ts` ...

hamzahsn: but let's also see what @Weetbix  think about it 😉

src/Sidebar/Sidebar.tsx
```diff
@@ -8,12 +7,37 @@ import {
...
+  href: string;
+};
+
+type HeaderProps = {
+  logoUrl: string;
+  logoAltText: string;
+  href: string;
+};
+
+interface LinkFC<T> extends React.FC<T> {
```
Weetbix, line 22: This is a bit over complicated. Display name is a react functional component feature, and so it exists on the FC type already.
Secondly, no need to use generics or extend the FC type here, you can use an intersection, and replace all this code with: 

```typescript
const Link: React.FC<LinkProps> = ({ 
  // .. 
})
Link.displayName = "Sidebar.Link";

const Header: React.FC<HeaderProps> = ({ 
  // .. 
})
Header.displayName = "Sidebar.Header";

interface SideBarSubComponents {
  Link: React.FC<LinkProps>;
  Header: React.FC<HeaderProps>;
}

const Sidebar: React.FC & SideBarSubComponents = ({ children }) => {
};

Sidebar.Header = Header;
Sidebar.Link = Link;
export default Sidebar;
```

definitions.d.ts
Weetbix, line 1: I'm a bit suspicious of this fix, as its more of a hack than an actual fix, and we lose the type information of the SVG: 
image

Can you try this solution: https://duncanleung.com/import-svg-storybook-webpack-loader/

src/Sidebar/sidebar.style.ts
```diff
@@ -31,7 +35,7 @@ export const SidebarHeaderStyles = styled.div`
   }
 `;
 
-export const LinkStyles = styled.div`
+export const LinkStyles = styled.div<ActiveProps>`
```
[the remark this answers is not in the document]
hamzahsn: I was going to use them like this but as I didnt want to replicate them so I typed it alone, but yeah, this one might be better :)

definitions.d.ts
hamzahsn: hmm but it's not actually a hack as using typescript with react when importing SVGs require declaring them as a module like this one: https://webpack.js.org/guides/typescript/#importing-other-assets

hamzahsn: and this one will do the fix for storybook and jsx when we need it, but I will try what you mentioned for sure as I remember that I saw it once before

Weetbix: Hmmm weird. Then we can update the definition to specify it as a string at least?

hamzahsn: it is actually how typescript works with svgs, and other undefined modules, so we declare them them like that but didnt see how to specify them as a string

Weetbix: Ok Lets go with this :)

PR #650, opened 2020-06-16
[ EBS - 132] - WEB | Convert Row component to Typescript

hamzahsn:
`Description:`

This PR is about converting Row component to typescript

`AC:`

out of scope:

`Jira Link:`

https://uvitadev.atlassian.net/browse/EBS-132

`Screenshot:`


src/Grid/Row/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+  column?: boolean;
+  fullWidth?: boolean;
+};
+
+const Row: FC<RowProps> = ({
+  children,
+  style = {},
+  space = "normal",
+  position = "start",
+  positionMd = null,
```
a7madgamal, line 28: does it need to be `null`? what happens if we just don't set a value?

hamzahsn: but is it a null value though

a7madgamal: if adding null adds no value please remove it because thats what `undefined` is for and in that case not setting any value is just simpler and less confusing

hamzahsn: alright Sir 😉

package.json
```diff
@@ -16,7 +16,7 @@
   ],
   "scripts": {
     "linter:fix": "prettier --write '{,!(node_modules|.git|storybook-static)/**/}*.{js,json,jsx,md}'",
-    "linter:check": "eslint src/**/* && eslint stories/*.jsx",
+    "linter:check": "eslint 'src/**/*.{tsx, jsx}' && eslint stories/*.jsx",
```
Weetbix, line 19: Can you explain this change?

It appears to remove linting for .js and .ts files

src/Grid/Row/index.tsx
```diff
@@ -0,0 +1,68 @@
+import React, { FC, CSSProperties, ReactNode } from "react";
+import Styles from "./index.style";
+import classNames from "../../utils/classNames";
+
+type RowProps = {
+  children: ReactNode;
```
Weetbix, line 6: Children doesnt need to be defined by us as it comes from the react Functional Component prop definition

src/Grid/Col/index.js
```diff
@@ -54,4 +55,5 @@ Col.defaultProps = {
   md: null,
   sm: null,
   xs: null,
+  center: null,
```
Weetbix, line 58: are these unrelated changes?

package.json
hamzahsn: remove linting ? 🤔 , so the test was failing cause it was not finding the path for components in the subfolders for `Col` and `Row` and thats why I did it like that, any suggestions ?

src/Grid/Row/index.tsx
hamzahsn: oops missed it 👀

src/Grid/Col/index.js
hamzahsn: this one was actually cause I have changed the `SideBar.jsx` to `SideBar.tsx` and there was a prop in the Col which is this one but not mentioned in the types so I added it here cause the next ticket is about changing sidebar to `tsx` in both story and component

package.json
Weetbix: Can you try again? previously this picks up all files and lints what it can including js and ts files. Here you are forcing it to only consider tsx and jsx files

src/Grid/Col/index.js
Weetbix: Ok so lets remove this change from this PR as its unrelated, you can add it back to the sidebar PR. Otherwise its difficult to review one chunk or work and say its working and makes sense as sits split across multiple PRs with reasonings existing for other unknown reasons

package.json
```diff
@@ -16,7 +16,7 @@
   ],
   "scripts": {
     "linter:fix": "prettier --write '{,!(node_modules|.git|storybook-static)/**/}*.{js,json,jsx,md}'",
-    "linter:check": "eslint src/**/* && eslint stories/*.jsx",
+    "linter:check": "eslint 'src/**/*.{tsx, jsx, ts, js}' && eslint stories/*.jsx",
```
Weetbix, line 19:
```suggestion
    "linter:check": "eslint 'src/**/*' && eslint stories/*.jsx",
```

src/Grid/Col/index.js
hamzahsn: already done in the last commits 😃

package.json
hamzahsn: this one will keep the error, no ? as it will not detect the files `Col`, `Row`, under Grid 🤔

Weetbix: I tried it and it works locally, but it picks up an extra issue with the index file you should look into. 

I think the difference is this includes quotes and passes the handling to eslint, the other one uses bash to expand and includes folders instead of files

src/Grid/Col/index.js
hamzahsn: ok, trying it

PR #651, opened 2020-06-16
[EBS - 133 ] - WEB | Convert Col component to Typescript

hamzahsn:
**Description:**

This PR is about converting Row component to typescript

**AC:**

**out of scope:**

**Jira Link:**

https://uvitadev.atlassian.net/browse/EBS-133

**Screenshot:**


src/Grid/Col/index.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { FC, CSSProperties, ReactNode } from "react";
+import classNames from "../../utils/classNames";
+
+type ColProps = {
+  children?: ReactNode;
```
a7madgamal, line 5: no need to type children unless we want to require it then lets remove the `?`

src/Grid/Col/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+  lg?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | "hide";
+  md?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | "hide";
+  sm?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | "hide";
+  xs?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | "hide";
+  offset?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11;
+  offsetMd?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11;
+  offsetSm?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11;
+  offsetXs?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11;
+  style?: CSSProperties;
+  center?: boolean;
```
a7madgamal, line 14: we don't use center lets remove it

hamzahsn: but `center` is used in storybook as a prop for `Col` component and that's why I typed it here

a7madgamal: you should remove it from there instead of adding it here without use

a7madgamal: or implement the center logic

hamzahsn: removing it from here and convert `Grid.jsx` to `Grid.tsx` will lead to a crush for the app cause center prop is not typed

hamzahsn: center logic ? 🤔

a7madgamal: thats why you should do one pr at a time.

src/Grid/Col/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+};
+
+const Col: FC<ColProps> = ({
+  children,
+  lg = 12,
+  md = null,
```
a7madgamal, line 19: please remove all `null`s

a7madgamal:
@hamzahsn @Weetbix I fixed eslint job and tested that it catches errors in correct files
had to use temporary ignore comments to avoid this pr getting bigger

PR #652, opened 2020-06-19
EBS-90 add missing value prop to Textarea

a7madgamal:
we need this for setting initial values to `Textarea` component


Weetbix:
Ah I think theres more to be done here than it looks. Can you create a ticket perhaps, as this component also does not implement the InputWrapper, and therefore doesnt work with validation.

PR #653, opened 2020-06-19
[ MAIN ] - fix imports

hamzahsn:
this is a small PR in order to fix the import in Sidebar

### DIFF
--- a/src/Sidebar/Sidebar.tsx
+++ b/src/Sidebar/Sidebar.tsx
@@ -1,5 +1,8 @@
 import React from "react";
-import { Row, Col, Icon } from "../index";
+import Row from "../Grid/Row";
+import Col from "../Grid/Col";
+import Icon from "../Icon";
+
 import {
   SidebarStyles,
   LinkStyles,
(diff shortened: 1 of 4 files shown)


hamzahsn:
> why do we have changes in unrelated snapshots?

which snapshots ?

a7madgamal:
@Weetbix does the snapshot changes make sense to you?

a7madgamal:
the issue is this: when we import anything through index our webpack needs to handle things like fonts etc and also I think we import so much unused code

a7madgamal:
sorry closed by mistake :D

Weetbix:
So can we write a test or a failing case for this? Why didn't storybook fail if the dist is broken for typescript?

a7madgamal:
I can look into an eslint rule, or we can fix imports in web-portal but I think both are technical tasks

hamzahsn:
as @a7madgamal  said that I think it's a trechnical task, and I think it can be fixed in the config for jest and/or storybook but it needs to take a look into that in a seperate investigation

a7madgamal:
@Weetbix the dist is fine, the other portal project needs to fix something in webpack I think

PR #661, opened 2020-06-22
EBS-149 Textarea component improvements

a7madgamal:
breaking changes to `Textarea`
removed: 
- `onInput` since `onChange` is clearer and matches most other components
- `required` is now optional to match `InputWrapper` usage, I think we should change it later to `required` since it's weird to mark optional fields with `*` IMO.

added: 
- `id`, `name`, `defaultValue`, `iconName`, `onChange`, `onBlur`, `validateOnChange` and `validateOnBlur`.

now uses the `ValidationContext` and InputWrapper.

image


.gitignore
```diff
@@ -13,8 +13,8 @@
 /public/dist/*.js.map
 
 # dist files
-/dist/*
-/storybook-static/*
+/dist
```
a7madgamal, line 16: just to color ignored folders in vscode properly :)

.circleci/config.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       - run: npm run linter:check
       - run:
           name: Tests
-          no_output_timeout: 15m # a workaround until we make it faster later
+          no_output_timeout: 30m # a workaround until we make it faster later
```
a7madgamal, line 41: the time increases with each new story 🤷

package.json
```diff
@@ -22,7 +22,7 @@
     "storybook:build": "build-storybook -s public",
     "build:package": "babel --delete-dir-on-start --extensions .ts,.tsx,.js,.jsx ./src --out-dir ./dist && tsc --project tsconfig.types.json --emitDeclarationOnly",
     "test": "npm run build:package && jest",
-    "test:update": "jest --updateSnapshot",
+    "test:update": "npm run build:package && npm run storybook:build && jest --updateSnapshot",
```
a7madgamal, line 25: this makes sure everything is ready before snapshots

Weetbix: nice one, thanks for fixing

tests/helpers.js
```diff
@@ -1,16 +1,11 @@
...
+  const location = path.join(__dirname, "..", "storybook-static");
+  console.log({ location });
 
-  if (fs.existsSync(location)) {
-    location = `file://${location}`;
-  } else {
-    location = "http://localhost:3006";
-  }
-  return location;
+  return `file://${location}`;
```
Weetbix, line 8: I think this change makes sense. We should only support one method of snapshotting that is identical everywhere

tests/helpers.js
```diff
@@ -1,16 +1,11 @@
 import path from "path";
-import fs from "fs";
 import devices from "puppeteer/DeviceDescriptors";
 
 export function getStorybookUrl() {
-  let location = path.join(__dirname, "..", "storybook-static");
+  const location = path.join(__dirname, "..", "storybook-static");
+  console.log({ location });
```
Weetbix, line 6: Is this intentional?

a7madgamal: lol nope

a7madgamal:
@Weetbix I check the css seems similar now

PR #662, opened 2020-06-23
Ebs 146 - Global Styles To TS

ghost:
### Refactoring Global Styles
Global Styles are converted to Typescript.

Since ShadowDom inherited from it's parent only a fraction of CSS Attributes the use of wrapper around the WebComponents is needed to use global styling. It is needed to break up the Global Styles to use them inside the extra WrapperStyle. It should keep it DRY and the wrapper itself receive it's content from a SSOT so to speak - Global Styles stay in the global style files.


src/GlobalStyles/global.style.ts
```diff
@@ -1,13 +1,15 @@
-import { createGlobalStyle } from "styled-components";
```
sblaettgen: It looks like I used squashed into the wrong direction, `global.style.ts `and `index.ts` shouldn't be behind that commit message.

Weetbix (approved):
Looks good!

Thanks for the changes :)

src/GlobalStyles/global.style.ts
```diff
@@ -0,0 +1,28 @@
+import { createGlobalStyle, css } from "styled-components";
+import { Colors } from "..";
```
Weetbix, line 2: Update to the full import path as discussed :) 

like sidebar:
`import Colors from "../Colors";`

Weetbix (changes requested):

src/Fonts/ShadowStyleFonts.ts
```diff
@@ -0,0 +1,15 @@
+import { css } from "styled-components";
+import DefaultFont from "./DefaultFont";
+
+// CSS Attributes which normally being inherited from html/body should
+// aren't used in ShadowDom. It's encapsulation prevents a natural inheritance.
+// All attributes which are also global in Web Components should be
+// added to the Shadow Styles
+const ShadowStylesFonts = css`
+  input,
+  textarea {
```
a7madgamal, line 10: does this affect input and textarea only?

src/Fonts/fonts.style.ts
```diff
@@ -0,0 +1,120 @@
+import { createGlobalStyle, css } from "styled-components";
+import DefaultFont from "./DefaultFont";
+import ShadowStylesFonts from "./ShadowStyleFonts";
+
+import ie9Compat00 from "../../public/fonts/TTNorms/35C396_0_0.eot";
```
a7madgamal, line 5: Nitpick, but it looks like lots of legacy stuff I hope we can get rid of some

src/Fonts/ShadowStyleFonts.ts
sblaettgen: Nice that you adressing this ... yes indeed.

It is a straight convert from the original file but it doesn't make sense for me either. It is the default font, which is also used in `body` ... therefore I would suggest to get rid of `input, textarea` and introduce him as default standard for everything.

src/Fonts/fonts.style.ts
sblaettgen: Yes, I just converted it and get it to work with the ShadowDOM. Also I wasn't aware at that moment that the library is full of deprecated legacy code. Maybe it is a job for someone who know's whats been used and what not to get rid of it.

a7madgamal (approved):
nice job, if you tested it on the other repo and it works as expected then lgtm :)

README.md
```diff
@@ -27,7 +27,9 @@ focused in scope while avoiding unrelated commits and changes.
 - Push your branch `git push origin cool.new.feature`
 - Submit a [pull request](https://github.com/VivyTeam/vivy-components/pulls) and add details about your changes
 - Add a reviewer (or ask someone in person) to go over your changes
-- If all goes well click the "Squash and Merge" button to close your PR (make sure to clean up and delete your branch)
+- When the PR is approved click the "Squash and Merge" button to close your PR (make sure to clean up and delete your branch)
+- Create a new release by following previous releases format and pick a relevant version according to the change (patch, minor, major)
```
a7madgamal, line 31: maybe mention github clearly or even better add this link https://github.com/vivyteam/vivy-components/releases

README.md
```diff
@@ -27,7 +27,9 @@ focused in scope while avoiding unrelated commits and changes.
...
+- If all goes well you should receive an email from npm that the publish went ok
```
a7madgamal, line 32: also here you can add this link for debugging if the job failed for some reason https://app.circleci.com/pipelines/github/VivyTeam/vivy-components?branch=master

Weetbix (approved):
Great thanks!

PR #663, opened 2020-06-23
[MAIN] bump dependencies

Weetbix:
This PR bumps most of our current dependencies, and I hope that it will help with the test timeouts. Notably here are jest and puppeteer and storybook bumps, although jest only up to version 25 as 26 is not yet supported by storybook. 

Running it locally it no longer seems to hang, and the CI passed on the first run.


a7madgamal:
next step: lower that ~12min run time. doesn't make sense to generate all of them hopefully we can only run related changes only

Weetbix:
@a7madgamal 

> nice!

Lets hope so lol!

Yeah, related changes would be great but I wonder if it's difficult to implement. 

I think there are some quick wins here as we are currently spending 6 mins on the build which is really high

PR #667, opened 2020-06-26
MAIN fix Textarea defaultValue

a7madgamal:
we were setting `value` instead of `defaultValue` which prevented changing the value


a7madgamal:
yeah tbh it's been ages since I used this approach I suggest we convert the form components to something more solid like https://github.com/tannerlinsley/react-form later

PR #681, opened 2020-07-08
[ESB-165] WEB | Convert the vivy-components Checkbox component to TS

fabwoofer:
Jira ticket: https://uvitadev.atlassian.net/browse/EBS-165

**What is this PR about?**
Converting the legacy checkbox component to typescript.

**AC**
* All checkbox related files have been converted to typescript
* No visual changes

**Out of scope**
We are mapping icon codes with a json file that is in the public folder, which I think is not the best place for this but the not-migrated Icon component is relying on it, so I think it makes sense to come up with a better solution in scope of the Icon component migration.

**Screenshot**
Screenshot 2020-07-08 at 14 49 35


tsconfig.types.json
```diff
@@ -6,8 +6,9 @@
     "skipLibCheck": true,
     "preserveConstEnums": true,
     "outDir": "dist",
-    "rootDir": "./src"
+    "rootDir": ".",
```
fabwoofer, line 9: I had to change the `rootDir` in order to include an import from the public folder. Might be worth changing back again once we are converting the Icon component and (maybe) change the way we are mapping the icon codes.

a7madgamal (approved):
👍

src/Checkbox/checkbox.style.ts
```diff
@@ -0,0 +1,57 @@
...
+const checkboxFilled = mapping["icon-checkbox-filled"];
+
+export const CheckBoxWrapper = styled.div`
+  display: flex;
+  align-items: center;
+`;
+
+export const CheckMark = styled.label`
+  padding: 0;
+  font-size: 24px;
```
sblaettgen, line 15: I don't know if we paid attention to the fact in other components in the past, don't look like it ... but could u change the px values to rem?
Also - we are using the 62.5% => 1.6rem trick in GlobalStyles so u can just change for example 24px to 2.4rem.

fabwoofer: I did discuss the lack of design tokens with @a7madgamal and I have been told that we only copy values for now?! I think it makes sense in general to create a system to re-use design tokens for fonts and spacing rather than re-writing styles. But not sure if this should be done in the scope of this ticket or if we should have a general cleanup?

fabwoofer: Also, did we implement the global styles trick also on the other repos that use the legacy version of this component? If not, then this would lead to a different font size than intended...

src/Checkbox/checkbox.style.ts
```diff
@@ -0,0 +1,57 @@
...
+  &:checked + ${CheckMark}:after {
+    font-family: "vivy-icons";
+    content: "\\${checkboxFilled}";
+  }
+`;
+
+export const Label = styled.label`
+  font-size: 16px;
+  line-height: 1.38;
+  letter-spacing: normal;
```
hamzahsn, line 39: hmm, do we need this `letter-spacing` ad I didnt see any changes  without it? 
Screenshot 2020-07-08 at 17 39 18

src/Checkbox/checkbox.style.ts
```diff
@@ -0,0 +1,57 @@
...
+`;
+
+export const MetaLabel = styled.label`
+  font-size: 14px;
+  font-weight: normal;
+  line-height: 1.29;
+  letter-spacing: normal;
```
hamzahsn, line 46: same comment as above

stories/Checkbox.story.tsx
```diff
@@ -1,14 +1,14 @@
 import React from "react";
 import { storiesOf } from "@storybook/react";
-import { Checkbox } from "../src/index";
+import { Checkbox } from "../src";
```
hamzahsn, line 3: 👍

hamzahsn (commented):
Good Job Fabian! but just dropped few comments :)

src/Checkbox/checkbox.style.ts
fabwoofer: This screenshot is not really useful since I don't have any point of reference. It might be that the letter spacing definition here is superfluous but then again, is this checkbox component used somewhere where we need this? Since this is a migration and somebody put this in on purpose, I don't see any harm in leaving this in.

tsconfig.storybook.json
```diff
@@ -3,7 +3,8 @@
     "esModuleInterop": true,
     "jsx": "react",
     "skipLibCheck": true,
-    "preserveConstEnums": true
+    "preserveConstEnums": true,
+    "resolveJsonModule": true
```
hamzahsn, line 7: forgot to ask, can't we declare `json` as a module in order for our TS to recognize it instead of adding TS rule ?

src/Checkbox/checkbox.style.ts
hamzahsn: hmm, yeah I had to make a gif actually but it's ok, let's leave it as it is if that ok, but the issue, cause I thought it's only affecting the label of the checkbox and that's why I found it useless

tsconfig.storybook.json
fabwoofer: It would also be possible to declare it in the type definition but I didn't see any gain in this. Could you elaborate on the benefits of your suggestion?

src/Checkbox/checkbox.style.ts
fabwoofer: Right, so after feedback from @sblaettgen and @hamzahsn I changed the size unit to REM. Will keep it in mind for further migrations, thanks guys 🙏

sblaettgen: @fabwoofer asked me about this in Slack ... it is just plain wrong to use `px` because the globals are setting everything up for REM even in the legacy code. My guess is that it was overlooked or rushed or whatever but they are (should -provided that the global styles are used correctly) even in legacy (JS) fully supported.

So I strongly voted for `rem` here ... it really is a bugfix 😄

ghost (approved):

tsconfig.storybook.json
hamzahsn: it is OK, with that actually, but the reason I mentioned that is that we can define modules in their own definition TS file and avoid adding multiple rules which makes things a bit confusing sometimes due to lots of rules(for me as example), and I had some kind of issue with this one using another lib before(e.g angular), but since TS v2.9 we can use what you did out there, so, all fine with that :)

src/Checkbox/checkbox.style.ts
hamzahsn: > Right, so after feedback from @sblaettgen and @hamzahsn I changed the size unit to REM. Will keep it in mind for further migrations, thanks guys 🙏

Actually we have to thank you for your great work 🙏  😄

fabwoofer:
I am having some issues with the rebase/re-signing of my unsigned commits, so I will close this PR and open another one to start fresh to not spend much more time on this

PR #686, opened 2020-07-15
Select Component.     JS -> TS

ghost:
The main task was done already, just fixed some import issues with the `Colors` and px -> rem


a7madgamal (approved):

fabwoofer:
Think you will need to update the tests since the snapshots changed

PR #821, opened 2022-08-04
MOG883: update design password-reset page

thebigJulien:
updating buttons design to vivy-next


amrwski (approved):

amrwski:
this change is meant to affect all buttons and not just the password-reset page, right?

thebigJulien:
> this change is meant to affect all buttons and not just the password-reset page, right?

@amrwski  I don't think another repo is using this one as dependency...

